Issue #5, opened 2016-11-09
Fix hourly dataframe creation


arbennett:
Currently the dataframe creation only creates up to midnight on the last day, but we need it to generate through the full day.

arbennett:
This can be accomplished via `metsim.stop + pd.Timedelta('1 days')`.

arbennett:
Solved via #10.

Issue #2, opened 2016-11-07
Move `mtclim.calc_solar_geom()`


arbennett:
The `calc_solar_geom()` function needs to be moved outside of `mtclim` so that it is clear that it is not _only_ a part of the `mtclim` calculations.  Values for this are calculated once for each grid cell, and used in both forcing generation as well as disaggregation.  It might be useful to just wrap this in a `solar` module or something similar, then call it before doing any forcing generation.

arbennett:
Done in #10.

Issue #1, opened 2016-11-07
Simplify global configuration storage


arbennett:
When initializing a run there are several places that could be laid out more intelligently.  The `options`, `metsim`, and `configuration` modules should be made more clear.  

The `options` module should simply handle command line stuff.  This could be moved to the main `ms` script quite easily.

The `metsim` and related `MetSim` objects should not hold any global information, but should defer to the `configuration` module.  It would then be required that if not running from the command line that the `configuration` module could be initialized before creating a `MetSim` object.  This means there are modifications that need to be made to the `configuration` module.

Edit: I will be appending small grained tasks that fall under this category as I find them.

* `metsim.start` and `metsim.end` are set in `options.init`.  This makes them confusing if you are just reading the code.

arbennett:
Solved via #10.

PR #10, opened 2016-11-18
Develop

arbennett:
This pull request addresses the following issues:

#1 - The configuration module is updated based on metadata that is read when reading the forcings.
#2 - The `Forcing` class has been created to manage the entire workflow, and now contains `calc_solar_geom`.
#5 - Hourly dataframe creation is fixed via the method discussed in the issue.

Additionally this pull request cleans up the entire structure of the code and adds example files to make running possible.  To run the code use `./ms -c examples/example.conf -n 1`.  An ascii output file will be generated in `./results/test_48_120`.  Although these are not the correct disaggregated forcings or the desired output format they make inspection easy and are illustrative of the entire process.


metsim/io.py
```diff
@@ -8,21 +8,21 @@
...
+    will be returned.
     """
     cfp = ConfigParser()
     if os.path.isfile(fpath):
         cfp.read(fpath)
     return cfp 
     
 
-def read_ascii(fpath, n_days=-1):
+def read_ascii(fpath, n_days=-1) -> Forcing:
```
jhamman: for this you can take a look here: https://github.com/jhamman/tonic/blob/master/tonic/models/vic/vic.py#L154

metsim/forcing.py
```diff
@@ -0,0 +1,161 @@
...
+from metsim.configuration import PARAMS as params
+from metsim.configuration import CONSTS as consts
+from metsim.configuration import OPTIONS as options
+
+class Forcing(object):
+    """
+    TODO
+    """
+    
+    def __init__(self, met_data: pd.DataFrame, params: dict):
```
jhamman: +1 on the type hints!!! - yay python 3.5!

metsim/io.py
```diff
@@ -35,12 +35,12 @@ def read_ascii(fpath, n_days=-1):
...
+    # TODO: FIXME: Finish this
+    return Forcing(None, None)
 
 
-def read_binary(fpath, n_days=-1):
-    """
-    TODO
-    """
+def read_binary(fpath: str, n_days=-1) -> Forcing:
+    """ Reads a binary forcing file (VIC 4 format) """
```
jhamman: nit: the pep8 docs don't have a space around the triple quotes in docstrings so I don't think they are necessary.

metsim/methods/mtclim.py
```diff
@@ -4,182 +4,103 @@
...
+
+    window = np.zeros(params['n_days'] + 90)
     df['t_max'] = np.maximum(df['t_max'], df['t_min'])
     dtr = df['t_max'] - df['t_min']
     sm_dtr = pd.rolling_window(dtr, window=30, freq='D', 
                                win_type=win_type).fillna(method='bfill')
-    if n_days <= 30:
+    if params['n_days'] <= 30:
         print('Timeseries is shorter than rolling mean window, filling ')
         print('missing values with unsmoothed data')
```
jhamman: use the warnings module here instead of print statements. That way we can filter them  when running this code over large domains...

metsim/io.py
bartnijssen: The more we can re-use existing pieces the better (if they do the right thing).

metsim/disaggregate.py
```diff
@@ -6,90 +6,103 @@
...
-tiny_rad_fract = np.zeros(366) #This is updated during the mtclim run
 
-def disaggregate(df_daily):
+def disaggregate(df_daily, solar_geom):
     """
     TODO
     """
-    dates_hourly = pd.date_range(metsim.start, metsim.stop, freq='H') 
+    end = metsim.stop + pd.Timedelta('1 days')
+    dates_hourly = pd.date_range(metsim.start, end, freq='H')
```
bartnijssen: So are we hard-coding the assumption that the final output will have a time resolution that is some integer multiple of hourly?

Since we allow VIC to be run at sub-hourly time steps, we should support that in the met generation as well. Even if we do not do so in the first implementation, then we need to comment on that and open an issue for extending this to sub-hourly

metsim/disaggregate.py
```diff
@@ -6,90 +6,103 @@
...
+    # temps = interp(range(len(df_hourly.index))
+    # return temps
+
+def precip(precip):
     """
-    TODO
+    Splits the daily precipitation evenly throughout the day
     """
-    pass
+    return precip.resample('H', how='sum').fillna(method='ffill')/24.
```
bartnijssen: Remove hard-coded literal (or use something like `hourspday`)

bartnijssen: What about time steps that are not hourly?

metsim/disaggregate.py
```diff
@@ -6,90 +6,103 @@
...
 
 
-def _disagg_thermal(df_daily, df_hourly):
+def longwave(longwave):
     """
-    TODO
+    Splits the daily longwave evenly throughout the day
     """
-    pass
+    return longwave.resample('H', how='sum').fillna(method='ffill')/24.
```
bartnijssen: Why not use `mean` or something like it for `how`? Same in the other cases.

metsim/disaggregate.py
```diff
@@ -6,90 +6,103 @@
...
 
 
-def _disagg_shortwave(df_daily, df_hourly):
+def shortwave(sw_rad, daylength, day_of_year, tiny_rad_fract):
     """
     TODO
     """
     tiny_step_per_hour = int(3600 / consts['SRADDT'])
-    tmp_rad = df_daily['s_swrad']
+    tmp_rad = sw_rad * daylength / 3600.
```
bartnijssen: remove hard-coded literal.

bartnijssen: Is `sw_rad` a cumulative flux or a mean flux? If the latter than this does not work.

metsim/disaggregate.py
```diff
@@ -6,90 +6,103 @@
...
-                    tinystep -= 24*tiny_step_per_hour
-                hourlyrad[i*24+j] += tiny_rad_fract[df_daily['day_of_year'][i]][tinystep]
-            #FIXME: This calculation is incorrect
-            hourlyrad[i*24+j] *= tmp_rad[i]
-    df_hourly['s_swrad'] = hourlyrad
-
-
-def set_min_max_hour(df_daily, df_hourly):
+    hourlyrad = np.zeros(n_days*24 + 1)
+    tiny_offset = (params.get("theta_l", 0) - params.get("theta_s", 0) / (24./360))
```
bartnijssen: remove hard-coded literals?

metsim/forcing.py
```diff
@@ -0,0 +1,158 @@
...
+        cosslp = np.cos(params['site_slope']  * consts['RADPERDEG'])
+        sinslp = np.sin(params['site_slope']  * consts['RADPERDEG'])
+        cosasp = np.cos(params['site_aspect'] * consts['RADPERDEG'])
+        sinasp = np.sin(params['site_aspect'] * consts['RADPERDEG'])
+        coszeh = np.cos(np.pi / 2.-(params['site_east_horiz'] * consts['RADPERDEG']))
+        coszwh = np.cos(np.pi / 2.-(params['site_west_horiz'] * consts['RADPERDEG']))
+        dt     = consts['SRADDT']  
+        dh     = dt / consts['SECPERRAD'] 
+    
+        tiny_step_per_day = 86400 / consts['SRADDT']
```
bartnijssen: remove hard-coded literals - in general throughout. Not going to repeat further

metsim/forcing.py
```diff
@@ -0,0 +1,158 @@
...
+        met_data['t_max']
+        params['lat']
+        params['lon']
+
+
+    def _calc_solar_geom(self):
+        """
+        Flat earth assumption
+        """
+        tt_max0 = np.zeros(366)
```
bartnijssen: remove hard-coded literals

metsim/disaggregate.py
arbennett: Would it be fine to assume the lowest time resolution of 1 minute?  If not, what should the minimum bound be?

jhamman: We should probably support `60s` resolution since that is what VIC supports.

metsim/disaggregate.py
jhamman: also, the new pandas syntax is `longwave.resample('H').sum().fillna(...`

metsim/io.py
bartnijssen: @arbennett : can you respond to Joe's comments and fix the small stuff. Let me know when that is done and I'll merge (tell me when it's ready)

metsim/disaggregate.py
arbennett: I made it so that timestep is specified in mintutes.  Maybe at a later date we can look at contextualizing between hours and minutes, but that's purely convenience so I'm not focusing on it currently.

metsim/disaggregate.py
arbennett: Fixed this one as well.

metsim/io.py
arbennett: Will implement this after the NetCDF stuff is in place.

metsim/io.py
arbennett: Done.  Everything should be ready in the current form.

jhamman:
@arbennett and @bartnijssen - 

I played around with this last week a bit, giving it a test run on a the sample data that is provided in the repo.  I have just a few general comments beyond the fact that I'm quite excited to see this so far along.

1. I'd like to see some tests. Perhaps those are forthcoming.
2. We should discuss code style. My vote would be for a fairly strict PEP8 style but I'll leave that up to you. I'd prefer for my editor not to blow up with style warnings when I open each metsim module (as it does currently).
3. As far as I could tell, we need to write out a few more variables by default in order to run VIC (I think humidity was the main one but a glance at the VIC documentation will be the best source there).

Issue #16, opened 2017-01-19
Longwave radiation is underestimated 


arbennett:
There is ~4% consistent underestimation of longwave radiation that I've been unable to track down fully.  Below is a plot showing this:  

![lw](https://cloud.githubusercontent.com/assets/7518706/22090519/0338fdb8-dda7-11e6-85b9-3eb041ab2174.png)

The red is the output of VIC4.2 and the blue underneath is the output of MetSim.  There are only 3 inputs to the longwave calculation: temperature, vapor pressure, and cloud fraction.  Both temperature and cloud fraction have deviations from VIC4.2 of <0.1%, and don't show a consistent bias direction for those deviations, so it seems they are not to blame.  Similarly, the vapor pressure doesn't show a consistent bias, so I don't think it is the cause either.  This leaves the calculation of the longwave radiation itself to be the problem.  The code to do so is:

```
def longwave(air_temp, vapor_pressure, tskc):
    """ Calculate longwave """
    emissivity_calc = {
            'DEFAULT'    : lambda x : x,
            'TVA'        : lambda x : 0.74 + 0.0049 * x,
            'ANDERSON'   : lambda x : 0.68 + 0.036 * np.power(x, 0.5),
            'BRUTSAERT'  : lambda x : 1.24 * np.power(x/air_temp, 0.14285714),
            'SATTERLUND' : lambda x : 1.08 * (1 - np.exp(-1 * np.power(x, (air_temp/2016)))),
            'IDSO'       : lambda x : 0.7 + 5.95e-5 * x * np.exp(1500/air_temp),
            'PRATA'      : lambda x : (1 - (1 + (46.5*x/air_temp)) *
                np.exp(-np.sqrt((1.2 + 3. * (46.5*x/air_temp)))))
            }
    cloud_calc = {
            'DEFAULT' : lambda x : (1.0 + (0.17 * tskc**2)) * x,
            'CLOUD_DEARDORFF' : lambda x : tskc + (1-tskc)*x
            }
    # Reindex and fill cloud cover, then convert temps to K
    tskc = tskc.reindex(air_temp.index)
    tskc = tskc.fillna(method='ffill')
    air_temp = air_temp + conf.KELVIN 

    # Calculate longwave radiation based on the options 
    emissivity_clear = emissivity_calc[conf.LW_TYPE.upper()](vapor_pressure)
    emissivity = cloud_calc[conf.LW_CLOUD.upper()](emissivity_clear) 
[...]

Issue #14, opened 2017-01-05
Fix vapor pressure / humidity calculation


arbennett:
The vapor pressure from mtclim (red) and metsim (black) are shown in the figure below.  At the beginning and end they match up nicely, but the middle is clearly an issue.  This section roughly corresponds to summer months.  

I am somewhat confident in the `tfmax` calculation, but unsure how.

![vp](https://cloud.githubusercontent.com/assets/7518706/21695885/7601f6a4-d340-11e6-917f-d6f85df844e9.png)

bartnijssen:
Can you put an actual date axis on the plot and units on the y-axis. I am not sure what I am looking at, so it is hard to comment. Also, where is this location supposed to be?

arbennett:
![vp](https://cloud.githubusercontent.com/assets/7518706/21789782/61b459f4-d68c-11e6-9b8e-35dfe99cd7b8.png)

This is taken from the Stehekin data.  The coordinates are (43.3125, -120.5625).

arbennett:
I have found that this is an error in the calculation of the effective annual precipitation.  I seem to have a factor of 10 left off somewhere.  The code I use to calculate this is:

    parray = np.array(pd.Series(window)
                  .rolling(window=90, win_type=win_type,axis=0)
                  .mean().fillna(method='bfill'))[90:] * consts['DAYS_PER_YEAR'] 
    parray = np.maximum(parray, 8.0)

I am not sure where this factor is coming from either in my code or in the VIC 4.2 code.

jhamman:
Is it possible this is just a units issue with the precipitation?  I'm guessing you are using VIC precipitation in mm/day whereas MTCLIM may be expecting cm/day ([ref](https://github.com/jhamman/mtclim/blob/master/mtclim43.c#L103)).

arbennett:
That seems like the most likely issue.  I'm not sure where in VIC this conversion is happening though, since I get the same disaggregated precipitation as MTCLIM without applying any unit conversions.

jhamman:
Right, mtclim never performed the precip disaggregation so that was handled by VIC. So, There was probably a multiply-by then divide-by thing going on within VIC.

arbennett:
As of the current commit to my fork the discrepancy looks like this:

![vp](https://cloud.githubusercontent.com/assets/7518706/22090457/81d6e79e-dda6-11e6-99b4-a5403119508f.png)

[...]

Issue #13, opened 2017-01-05
Fix endpoints of temperature disaggregation


arbennett:
Shown below are the beginning and end (respectively) of a MetSim run (black) and a MTCLIM run (red).  As can be seen, they match quite well, except for the very ends.  This is because we don't provide the right endpoints.  To calcluate them for the figures shown I was using `mean(tmin, tmax)` for the first and last days.  This is clearly incorrect, but I haven't looked into what I should be using.

## Beginning
![temp_begin](https://cloud.githubusercontent.com/assets/7518706/21669553/ac24f5da-d2c1-11e6-8696-a4cafa8d7da9.png)

## End
![temp_end](https://cloud.githubusercontent.com/assets/7518706/21669555/af9c2b2a-d2c1-11e6-8e14-ddc8b1297800.png)

arbennett:
This is done in VIC by taking the 2nd value and 2nd to last value and repeating them.  Taking this approach in MetSim fixes the issue.

bartnijssen:
There still seems to be a time offset though? Any idea where that originates? Same comment as for issue #14 : Please use time axis on x-axis and units on y-axis.

arbennett:
![temp](https://cloud.githubusercontent.com/assets/7518706/21790209/7a2e4924-d68f-11e6-83be-5b566d84a2cd.png)

That was an indexing issue that came up because the first point wasn't supplied, so everything got shifted back.  Here's an updated one so you can see.

bartnijssen:
@arbennett : If this is fixed go ahead and close the issue. It makes it easier to quickly check what is still outstanding for the first release

arbennett:
I've left it open since it's only fixed in my fork.  When I open the pull request I'll reference all of these issues that it addresses, and when it's finally merged I'll close them all.

bartnijssen:
I much prefer many smaller PRs that roughly correspond to individual issues rather than a single large PR at the end. They are hard to review, which means that no one actually reviews them

arbennett:
Fixed in #17

Issue #3, opened 2016-11-07
Speed up `calc_solar_geom()`


arbennett:
Currently the `calc_solar_geom()` function takes approximately 85% of the runtime of MTCLIM.  It seems like it should be easy to lower this runtime quite significantly due to two large loops.

arbennett:
Improved in #10, but I would still like to leave this open because there's room for further improvement.  This portion of the calculation still take ~40% of the runtime of MetSim.

jhamman:
Something to look at is @klapo's [solargeo](https://github.com/klapo/solargeo). It is a different (@klapo would say more accurate) way to calculate solar geometry. He optimized it a bit with Cython.

bartnijssen:
Let's not worry too much about this for now. I'd rather have a working version that reproduces earlier results before we dive into this. At this point this falls under the heading "premature optimization" since the choice of solar geometry calculation does not affect the overall structure of the code.

klapo:
Just for your reference, the current version of MTCLIM doesn't calculate orbital parameters correctly. It puts perihelion and aphelion at the wrong time of year. I suspect there is some compensating going on since MTCLIM performs well overall.

The orbital parameters also give you a bad sub-daily disaggregation, especially at longer time steps. This largely doesn't cause issues until you try to distribute solar radiation at a sub-daily time step onto complex terrain.

bartnijssen:
@klapo : Thanks but not our main concern at the moment. I first want a version of MetSim that reproduces `mtclim` as implemented in VIC and then make improvements to it. At that time the improved solar geometry would be great to include.

arbennett:
Addressed in #17

PR #17, opened 2017-01-21
Develop

arbennett:
This pull request addresses issues #3, #13, #14, and #16.  At this point the code reproduces to within <1% error what VIC4.2 does when used to generate forcings.  It additionally supports NetCDF input and output and is made extensible by compartmentalizing the core functions: solar geometry calculation, forcing generation (currently implemented as MTCLIM), and disaggregation.

Once merged, remaining work for 1.0.0 release will include improvements to documentation, tests which show that this code reproduces VIC4.2, bugfixes, and adding options such as verbose mode and to re-hook up parallelism.


metsim/disaggregate.py
```diff
@@ -7,107 +7,179 @@
...
+    return lwrad, tskc
+
+
+def shortwave(sw_rad, daylength, day_of_year, tiny_rad_fract, params):
     """
     TODO
     """
-    tiny_step_per_hour = int(consts['SEC_PER_HOUR'] / consts['SRADDT'])
-    tmp_rad = sw_rad * daylength / consts['SEC_PER_HOUR'] 
+    tiny_step_per_hour = cnst.SEC_PER_HOUR / cnst.SRADDT
```
orianac: Why don't you have to do `int()` anymore?

metsim/disaggregate.py
```diff
@@ -7,107 +7,179 @@
...
+    disaggrad = np.zeros(int(n_days*ts_per_day) + 1)
+    tiny_offset = ((params.get("theta_l", 0) - params.get("theta_s", 0) 
+                   / (cnst.HOURS_PER_DAY/360)))
+    
     # Tinystep represents a daily set of values - but is constant across days
-    tinystep= np.arange(consts['HOURS_PER_DAY'] * tiny_step_per_hour) - tiny_offset
-    tinystep[np.array(tinystep<0)] += consts['HOURS_PER_DAY'] * tiny_step_per_hour
-    tinystep[np.array(tinystep>(24*tiny_step_per_hour-1))] -= consts['HOURS_PER_DAY'] * tiny_step_per_hour
-    chunk_sum = lambda x : np.sum(x.reshape((len(x)/120, 120)), axis=1)
+    tinystep= np.arange(cnst.HOURS_PER_DAY * tiny_step_per_hour) - tiny_offset
```
orianac: add a space after `tinystep`

metsim/methods/mtclim.py
```diff
@@ -26,74 +21,61 @@ def run(forcing: pd.DataFrame, disagg=True):
     lat_idx = forcing.index.names.index('lat')
     time_idx = forcing.index.names.index('time')
     sg = solar_geom(forcing['elev'][0], forcing.index.levels[lat_idx][0])
+    sg = {'tiny_rad_fract' : sg[0],
```
orianac: Is it safe to overwrite the variable `sg` while drawing from it?

metsim/methods/mtclim.py
```diff
@@ -26,74 +21,61 @@ def run(forcing: pd.DataFrame, disagg=True):
...
-    dZ = (params['site_elev'] - params['base_elev'])/1000.
+    dZ = (df['elev'][0] - params['base_elev'])/1000.
     lapse_rates = [params['t_min_lr'], params['t_max_lr']]
-    df['t_max'] = df['t_max'] + dZ * lapse_rates[1]
-    df['t_min'] = np.minimum(df['t_min'] + dZ * lapse_rates[0], df['t_max']-0.5)
-    t_mean = np.mean(df['t_min'] + df['t_max'])
-    df['t_day'] = ((df['t_max'] - t_mean) * params['TDAYCOEF']) + t_mean
+    t_max = df['t_max'] + dZ * lapse_rates[1]
+    t_min = np.minimum(df['t_min'] + dZ * lapse_rates[0], t_max-0.5)
+    t_mean = (t_min + t_max)/2
```
orianac: Why `/2` instead of mean?

metsim/methods/mtclim.py
```diff
@@ -106,127 +88,90 @@ def _calc_tfmax(precip, dtr, sm_dtr):
...
-            if end_yday == start_yday-1:
-                isloop = True
-        else:
-            if end_yday == 365 or end_yday == 366:
-                isloop = True
-        if isloop:
-            for i in range(90):
-                window[i] = df['precip'][params['n_days'] - 90 + i]
+        start_day, end_day = df['day_of_year'][0], df['day_of_year'][-1]
+        if (start_day%365 == (end_day%365)+1) or (start_day%366 == (end_day%366)+1):
```
orianac: nifty

metsim/methods/mtclim.py
```diff
@@ -106,127 +88,90 @@ def _calc_tfmax(precip, dtr, sm_dtr):
...
     potrad = sg['potrad']
     daylength = sg['daylength']
     yday = df['day_of_year'] - 1
 
-    # Calculation of shortwave is split into 3 components:
-    # 1. Direct radiation arriving from incident light
-    # 2. Diffuse radiation over entire daylenght
-    # 3. Influence of snowpack - optionally set by MTCLIM_SWE_CORR
-    t_tmax = np.maximum(tt_max0[yday] + (params['ABASE'] * pva), 0.0001)
+    t_tmax = np.maximum(tt_max0[yday] + (cnst.ABASE * pva), 0.0001)
```
orianac: Why take out the documentation? Or do you no longer do those steps?

orianac: OH you moved it. I see.

metsim/physics.py
```diff
@@ -161,84 +162,85 @@ def svp_slope(temp, a=0.61078, b=17.269, c=237.3):
...
+
+
+@jit(nopython=True, cache=True)
+def solar_geom(elev, lat):
     """
     Flat earth assumption
     """
-    dayperyear = int(np.ceil(consts['DAYS_PER_YEAR']))
+    # optical airmass by degrees
+    OPTAM = [2.90,  3.05,  3.21,  3.39,  3.69,  3.82,  4.07,
```
orianac: Is there a citation for this?

examples/example.conf
```diff
@@ -5,16 +5,16 @@
...
+start = 1949/1/1:0  
 
 # Forcings end at this date (year/month/day)
-stop = 1950/1/31
+stop = 1949/12/31
 
 # Input and output directories
-forcing = ./examples/forcing_sub.nc
-domain  = ./examples/domain_sub.nc
+forcing = ./examples/forcing
```
arbennett: That may make it clearer.  This was a carryover from the example VIC data I have been using, which didn't have an extension.

metsim/disaggregate.py
```diff
@@ -7,107 +7,179 @@
...
-                                       solar_geom['tiny_rad_fract']))
-    df_disagg['temp'] = temp(df_daily, df_disagg)
-    df_disagg['precip'] = precip(df_daily['precip'])
-    df_disagg['longwave'] = longwave(df_daily['lwrad'])
-    df_disagg['wind'] = wind(df_daily['wind'])
+                                        df_daily['dayl'],
+                                        df_daily['day_of_year'],
+                                        solar_geom['tiny_rad_fract'],
+                                        params))
+
```
arbennett: Each call to a function in the `disaggregate` module will return the disaggregated values for the variable named by the function.   Documentation is a huge issue with this codebase right now, I'll admit.

metsim/disaggregate.py
```diff
@@ -7,107 +7,179 @@
...
     TODO
     """
-    ts = float(params['time_step'])
     rad_mask = 1*(disagg_rad > 0)
     diff_mask = np.diff(rad_mask)
     rise_times = np.where(diff_mask>0)[0] * ts 
     set_times = np.where(diff_mask<0)[0] * ts
     t_Tmax = 0.67 * (set_times - rise_times) + rise_times
-    t_Tmin = rise_times - float(params['time_step'])
+    t_Tmin = rise_times
```
arbennett: This is intended.  In the VIC code this offset was necessary after a bunch conditionals, but we've implicitly built this into the `np.diff` call.

metsim/disaggregate.py
```diff
@@ -7,107 +7,179 @@
...
-
+    return 100. * np.minimum(vapor_pressure*1000/svp(temp), 1.0)
+   
 
-def shortwave(sw_rad, daylength, day_of_year, tiny_rad_fract):
+def vapor_pressure(hum_daily, temp, t_Tmin, n_out, ts):
+    """Calculate vapor pressure"""
+    # Scale down to milibar
+    vp_daily = hum_daily / 1000
+    # Linearly interpoolate the values
```
arbennett: Freudian slip, will fix.

metsim/disaggregate.py
arbennett: It wasn't necessary.  My response is, why did I even do that to begin with?

metsim/methods/mtclim.py
arbennett: The call to `np.mean` collapses that down to a single value unless you provide it an `axis` argument.  I think this is more clear, and less prone to error in the future, while it doesn't affect performance in any significant (or measurable) way.  This was a bug that eluded me for quite a while, unfortunately.

metsim/physics.py
arbennett: There is not one in `mtclim_vic.c` in the VIC code surrounding the definition, but maybe within other documentation somewhere.

metsim/disaggregate.py
```diff
@@ -7,107 +7,179 @@
...
+    ts_ends = cnst.MIN_PER_HOUR * cnst.HOURS_PER_DAY
+    time = np.append(np.insert(time, 0, time[0:2]-ts_ends), time[-2:]+ts_ends)
+    temp = np.append(np.insert(temp, 0, temp[0:2]), temp[-2:])
     
+    # Interpolate the values
     try:
         interp = scipy.interpolate.PchipInterpolator(time, temp, extrapolate=True)
-        temps = interp(float(params['time_step']) * np.arange(0, len(df_disagg.index)))
+        temps = interp(ts * np.arange(0, len(df_disagg.index)))
     except ValueError:
```
arbennett: Good question... this may not even be the long term desired behavior.  The call can fail when the call to `PchipInterpolator` is given a timeseries that is not strictly increasing, or the `time` and `temp`  arguments have mismatched lengths.

metsim/methods/mtclim.py
arbennett: Yup, it's the same principle around variable swapping (eg `a, b = b, a`)  In Python the right hand side is evaluated before the assignment is made.

metsim/disaggregate.py
```diff
@@ -7,107 +7,179 @@
...
+                   / (cnst.HOURS_PER_DAY/360)))
+    
     # Tinystep represents a daily set of values - but is constant across days
-    tinystep= np.arange(consts['HOURS_PER_DAY'] * tiny_step_per_hour) - tiny_offset
-    tinystep[np.array(tinystep<0)] += consts['HOURS_PER_DAY'] * tiny_step_per_hour
-    tinystep[np.array(tinystep>(24*tiny_step_per_hour-1))] -= consts['HOURS_PER_DAY'] * tiny_step_per_hour
-    chunk_sum = lambda x : np.sum(x.reshape((len(x)/120, 120)), axis=1)
+    tinystep= np.arange(cnst.HOURS_PER_DAY * tiny_step_per_hour) - tiny_offset
+    tinystep[np.array(tinystep<0)] += cnst.HOURS_PER_DAY * tiny_step_per_hour
+    tinystep[np.array(tinystep>(24*tiny_step_per_hour-1))] -= (
```
jhamman: I don't think you need the `np.array()` in either of these lines.

metsim/methods/mtclim.py
jhamman: should we just have `solor_geom` return a dictionary instead?

metsim/methods/mtclim.py
```diff
@@ -26,74 +21,61 @@ def run(forcing: pd.DataFrame, disagg=True):
...
     """
-    dZ = (params['site_elev'] - params['base_elev'])/1000.
+    dZ = (df['elev'][0] - params['base_elev'])/1000.
     lapse_rates = [params['t_min_lr'], params['t_max_lr']]
-    df['t_max'] = df['t_max'] + dZ * lapse_rates[1]
-    df['t_min'] = np.minimum(df['t_min'] + dZ * lapse_rates[0], df['t_max']-0.5)
-    t_mean = np.mean(df['t_min'] + df['t_max'])
-    df['t_day'] = ((df['t_max'] - t_mean) * params['TDAYCOEF']) + t_mean
+    t_max = df['t_max'] + dZ * lapse_rates[1]
+    t_min = np.minimum(df['t_min'] + dZ * lapse_rates[0], t_max-0.5)
```
jhamman: see comments about using the pandas where method above.

metsim/methods/mtclim.py
```diff
@@ -26,74 +21,61 @@ def run(forcing: pd.DataFrame, disagg=True):
...
-def calc_precip(df: pd.DataFrame):
+def calc_precip(df: pd.DataFrame, params: dict):
     """ Adjust precipitation according to isoh """
     df['precip'] = (df['precip'] * (df.get('site_isoh', 1) / df.get('base_isoh', 1)))
 
 
-def calc_snowpack(df: pd.DataFrame):
+def calc_snowpack(df: pd.DataFrame, params: dict, snowpack=0.0):
     """Calculate snowpack as swe."""
+    swe = np.ones(params['n_days']) * snowpack
```
jhamman:
```Python
swe = pd.Series(np.ones(params['n_days']), index=df.index) * snowpack
```

metsim/methods/mtclim.py
```diff
@@ -26,74 +21,61 @@ def run(forcing: pd.DataFrame, disagg=True):
...
+    accum = np.array(df['t_min'] <= cnst.SNOW_TCRIT)
+    melt = np.array(df['t_min'] >  cnst.SNOW_TCRIT)
```
jhamman: No need to cast to numpy arrays.

metsim/methods/mtclim.py
```diff
@@ -106,127 +88,90 @@ def _calc_tfmax(precip, dtr, sm_dtr):
...
-        for i in range(params['n_days']):
-            sum_precip = 0.0
-            for j in range(90):
-                sum_precip += window[i+j]
-                sum_precip = (sum_precip / 90.) * consts['DAYS_PER_YEAR']
-            sum_precip = np.maximum(sum_precip, 8.0)
-            parray[i] = sum_precip
+        parray = np.array(pd.Series(window)
+                            .rolling(window=90, win_type=win_type,axis=0)
+                            .mean())[90:] * cnst.DAYS_PER_YEAR
```
jhamman: 2 pep comments: 1) space after `,`, 2) pep8 prefers the break after a `(` rather than at the period. In either event, the indentation is off here.

I think it also makes sense to keep parray as a Series.  Also, also, just fyi, anytime you need a numpy array out of a Series, just use the `.values` attribute.

metsim/methods/mtclim.py
```diff
@@ -106,127 +88,90 @@ def _calc_tfmax(precip, dtr, sm_dtr):
...
+        eff_ann_precip = (sum_precip / params['n_days']) * cnst.DAYS_PER_YEAR
         eff_ann_precip = np.maximum(eff_ann_precip, 8.0)
         parray = eff_ann_precip
     else:
         # Calculate effective annual precip using 3 month moving window
-        parray = np.zeros(params['n_days'])
-        start_yday = df['day_of_year'][0]
-        end_yday = df['day_of_year'][-1]
-
+        window = np.zeros(params['n_days'] + 90)
```
jhamman: window should be a `Pandas.Series`

metsim/methods/mtclim.py
```diff
@@ -106,127 +88,90 @@ def _calc_tfmax(precip, dtr, sm_dtr):
...
-            for j in range(90):
-                sum_precip += window[i+j]
-                sum_precip = (sum_precip / 90.) * consts['DAYS_PER_YEAR']
-            sum_precip = np.maximum(sum_precip, 8.0)
-            parray[i] = sum_precip
+        parray = np.array(pd.Series(window)
+                            .rolling(window=90, win_type=win_type,axis=0)
+                            .mean())[90:] * cnst.DAYS_PER_YEAR 
+    # Convert to mm 
+    parray = np.maximum(parray, 80.0) / 10
```
jhamman:
```Python
parray = parray.where(parray < 80, 80) / 10
```

Replace both 80 and 10 with constants.

metsim/methods/mtclim.py
```diff
@@ -106,127 +88,90 @@ def _calc_tfmax(precip, dtr, sm_dtr):
...
 
     df['tfmax'] = _calc_tfmax(df['precip'], dtr, sm_dtr) 
-
     tdew = df.get('tdew', df['t_min'])
     pva = df.get('hum', svp(tdew))
-    pa = atm_pres(params['site_elev'])
+    pa = atm_pres(df['elev'][0])
+    #pa = atm_pres(5688)
```
jhamman: remove commented out code.

metsim/methods/mtclim.py
```diff
@@ -106,127 +88,90 @@ def _calc_tfmax(precip, dtr, sm_dtr):
...
     yday = df['day_of_year'] - 1
```
jhamman: The `day_of_year` series may not be necessary since this could also come from the index itself (e.g. `df.index.dayofyear`)

metsim/methods/mtclim.py
jhamman: see comments above on sticking with pandas whenever possible and using `.where`

metsim/metsim.py
```diff
@@ -18,37 +18,61 @@ class MetSim(object):
...
         """
-        # Record parameters and broadcast to the config module
-        self.params = params
-        configuration.update(params) 
+        # Record parameters 
+        self.update(params)
 
         # Get the necessary information from the domain 
-        domain = Dataset(configuration.PARAMS['domain'], 'r')
+        domain = Dataset(MetSim.params.get('domain', None), 'r')
```
jhamman: won't `Dataset` raise an error if it gets `None`? I would just let python raise a `KeyError` with `Dataset(MetSim.params['domain'], 'r')` so as not to confuse the error.

metsim/metsim.py
```diff
@@ -18,37 +18,61 @@ class MetSim(object):
...
         # Where we will store the results for writing out
         out_dict = {}
-
+        method = MetSim.methods[MetSim.params.get('method', 'mtclim')]
+           
         # Coerce the forcings into a list if it's not one
         if type(forcings) is not list:
             forcings = [forcings]
-
+
```
jhamman: remove extra whitespace

metsim/metsim.py
```diff
@@ -57,14 +81,12 @@ def run(self, method, forcings):
                 for j in range(len(met_data.lon)):
                     out_dict[i][j] = method.run(
                                 met_data.isel(lat=[i], lon=[j])
-                                .to_dataframe(), disagg=True)
-                    # TODO: Remove this
-                    print(i,j) 
+                                .to_dataframe(), MetSim.params, disagg=True)
```
jhamman: see comment above re line breaks

metsim/metsim.py
```diff
@@ -82,50 +109,57 @@ def write(self, data: dict):
...
                 }
-        dispatch[self.params.get('out_format', 'netcdf')](data)
+        dispatch[MetSim.params.get('out_format', 'netcdf').lower()](data)
          
 
     def init_netcdf(self):
         """Initialize the output file"""
         print("Initializing netcdf...")
-        out_file = os.path.join(self.params['out_dir'], "forcing.nc")
+        out_file = os.path.join(MetSim.params.get('out_dir', './results/'), "forcing.nc")
```
jhamman: since you're inferring the output path, you need to make sure it exists (`makedirs`)

metsim/metsim.py
```diff
@@ -82,50 +109,57 @@ def write(self, data: dict):
...
         self.output = Dataset(out_file, 'w')
 
         # Number of timesteps 
-        n_ts = len(pd.date_range(self.params['start'], 
-                self.params['stop'] + pd.Timedelta('1 days'), 
-                freq=self.params['time_step']+'T'))
+        n_ts = len(pd.date_range(MetSim.params['start'], 
+                MetSim.params['stop'] + pd.Timedelta('1 days'), 
+                freq=MetSim.params['time_step']+'T'))
```
jhamman: use proper string formatting here and fix the indentation.

metsim/metsim.py
```diff
@@ -82,50 +109,57 @@ def write(self, data: dict):
...
         
         # Create dimensions
         self.output.createDimension('lat', len(self.lats))
         self.output.createDimension('lon', len(self.lons))
         self.output.createDimension('time', n_ts) 
 
         # Create output variables
-        for varname in self.params['out_vars']:
+        for varname in MetSim.params.get('out_vars', []):
```
jhamman: I think it would make sense to default to the big seven variables.

metsim/metsim.py
```diff
@@ -82,50 +109,57 @@ def write(self, data: dict):
...
-        """Write out to the output file"""
+        """Write out as NetCDF to the output file"""
         print("Writing netcdf...")
         lats = data.keys()
         for lat in lats:
             lons = data[lat].keys()
             for lon in lons:
-                for varname in self.params['out_vars']:
-                    self.output.variables[varname][:, lat, lon] = data[lat][lon][varname].values
+                for varname in MetSim.params.get('out_vars', []):
```
jhamman: same as above.

metsim/metsim.py
```diff
@@ -82,50 +109,57 @@ def write(self, data: dict):
...
+        """Write out as ASCII to the output file"""
         print("Writing ascii...")
-        raise NotImplementedError 
-   
+        out_base = os.path.join(MetSim.params['out_dir'], "forcing")
+        lats = data.keys()
+        for lat in lats:
+            lons = data[lat].keys()
+            for lon in lons:
+                print(lat, lon)
```
jhamman: since this could end up being quite long, probably better to use logging module.

ms
```diff
@@ -61,24 +59,25 @@ def init(opts):
...
+             "out_vars"   : to_list(config['out_vars']),
+             "time_step"  : config['time_step'],
+             "out_format" : config['out_format']}
     return setup 
 
 
 def main():
     """Runs MetSim"""
     from metsim.metsim import MetSim 
     setup = init(parse(sys.argv[1:]))
```
jhamman: can we use configparser instead of `sys.argv`?

metsim/methods/mtclim.py
arbennett: I'm unclear here why we would want to do this over the current behavior.

metsim/disaggregate.py
```diff
@@ -7,107 +7,179 @@
...
 
-def wind(wind):
+def relative_humidity(vapor_pressure, temp):
     """
-    Wind is assumed constant throughout the day
+    TODO
     """
-    return wind.resample(params['time_step']+'T').fillna(method='ffill')
-
+    return 100. * np.minimum(vapor_pressure*1000/svp(temp), 1.0)
```
arbennett: I'm having a hard time thinking of a way that switching the `1000` our for a constant would improve the code.  Naming it `THOUSAND` seems really pedantic, and we can't just call it `CONVERSION_FACTOR` since that's not descriptive enough.  Doing something like `KILO` could work, but realistically that's exactly the same as putting `THOUSAND`.  What would you suggest putting for these?

metsim/disaggregate.py
```diff
@@ -7,107 +7,179 @@
...
+        'SATTERLUND' : lambda x : 1.08 * (
+            1 - np.exp(-1 * np.power(x, (air_temp/2016)))),
+        'IDSO'       : lambda x : 0.7 + 5.95e-5 * x * np.exp(1500/air_temp),
+        'PRATA'      : lambda x : (1 - (1 + (46.5*x/air_temp)) *
+            np.exp(-np.sqrt((1.2 + 3. * (46.5*x/air_temp)))))
+        }
+    cloud_calc = {
+        'DEFAULT' : lambda x : (1.0 + (0.17 * tskc**2)) * x,
+        'CLOUD_DEARDORFF' : lambda x : tskc + (1-tskc)*x
+        }
```
arbennett: On 2: I just ran the checks and there's no discernible difference in speed for ~500 runs.  I'm not sure about memory, but if it becomes an issue in the future we can change it.

metsim/disaggregate.py
```diff
@@ -7,107 +7,179 @@
...
+    # Account for end points
+    ts_ends = cnst.MIN_PER_HOUR * cnst.HOURS_PER_DAY
+    time = np.append(np.insert(time, 0, time[0:2]-ts_ends), time[-2:]+ts_ends)
+    temp = np.append(np.insert(temp, 0, temp[0:2]), temp[-2:])
     
+    # Interpolate the values
     try:
         interp = scipy.interpolate.PchipInterpolator(time, temp, extrapolate=True)
-        temps = interp(float(params['time_step']) * np.arange(0, len(df_disagg.index)))
+        temps = interp(ts * np.arange(0, len(df_disagg.index)))
```
arbennett: It looks like it is possible, but we'd have to first build the series with the `temp` data in the right location, and then call `.interpolate(method="pchip")` which will just call the `scipy` function here anyways.

metsim/methods/mtclim.py
arbennett: Unfortunately, we can't do that since that forces numba into object mode, which drastically slows this down.

metsim/methods/mtclim.py
arbennett: As it stands they do need to be cast into numpy arrays.  We could change this to use something like `np.select` but I think that this is more readable.  I'd like to see a better way of doing this whole thing, but this was the most concise and readable that I could come up with.

metsim/metsim.py
arbennett: Ooops, this is done in the executable right now, but that would definitely cause issues if someone was trying to script with this.  Fix incoming.

metsim/methods/mtclim.py
jhamman: Mainly, this would allow you to stick with pandas. I think it could also be written as

```Python
swe = pd.Series(snowpack, index=df.index)
```

Hard to argue that is less concise.

metsim/methods/mtclim.py
jhamman: What exactly are you trying to do?

How are these lines any different:

```Python
accum = np.array(df['t_min'] <= cnst.SNOW_TCRIT)
```
and 
```Python
accum = (df['t_min'] <= cnst.SNOW_TCRIT)
```

metsim/disaggregate.py
jhamman: It may feel overly verbose but it makes the code much easier if you name the conversion factor something meaningful, for example:

```Python
MM_PER_M = 1000.
PA_PER_KPA = 1000.
```
Then, when you read the code, it is immediately clear why/how you are changing the units.

metsim/disaggregate.py
```diff
@@ -7,107 +7,171 @@
...
-        temps = np.full(len(df_disagg.index), np.nan)
-
+    # Account for end points
+    ts_ends = cnst.MIN_PER_HOUR * cnst.HOURS_PER_DAY
+    time = np.append(np.insert(time, 0, time[0:2]-ts_ends), time[-2:]+ts_ends)
+    temp = np.append(np.insert(temp, 0, temp[0:2]), temp[-2:])
+
+    # Interpolate the values
+    interp = scipy.interpolate.PchipInterpolator(time, temp, extrapolate=True)
+    temps = interp(ts * np.arange(0, len(df_disagg.index)))
```
jhamman: I'm coming back to this because I really don't like it.  There are two main downsides to using scipy's function instead of pandas. First, it means we can only use one method of interpolation (pchip). I'm not saying we want to use other methods, but it would be nice to allow that flexibility. The second reason is that you use two lines to do what pandas would do with one. It's not any clearer to read and you end up with a bunch of logic that pandas already does for you. Let's not recreate the wheel.

metsim/methods/mtclim.py
jhamman: okay, let's note that in a comment though because some punk is going to come along one day and see this and "clean it up" unknowingly.

metsim/metsim.py
```diff
@@ -55,87 +81,100 @@ def run(self, method, forcings):
...
-        """ TODO """
+        """Write out as ASCII to the output file"""
         print("Writing ascii...")
-        raise NotImplementedError 
-   
+        out_base = os.path.join(MetSim.params['out_dir'], "forcing")
+        lats = data.keys()
+        for lat in lats:
+            lons = data[lat].keys()
+            for lon in lons:
```
jhamman: line 160 isn't needed.

```Python
for lon in data[lat]:
    ....
```
would do the same thing.

jhamman:
I looked through the changes and things are looking better. I still have some concerns about the under-use of pandas objects / over-use of numpy arrays. @arbennett - it seems like you are hesitant to rely on pandas timeseries index alignment and higher level tools. My view on this is that by using pandas as much as possible here, we can eliminate many potential bugs that may arise in future development of MetSim, not to mention leveraging tools that have already been implemented for this sort of task.

My other comment is that we should run this through a linter. I try not to be too much of a stickler on PEP8'ing everyone's code but in reality, standardized code styles make code easier to read. As it stands right now, there are at least a few dozen style errors that should be addressed.

metsim/methods/mtclim.py
arbennett: One of them is a Pandas Series and one of them is a numpy array.  If `swe` is a numpy array you can't index on a Series.  I'm moving it to a Series per your request, you'll see the fix soon.

metsim/methods/mtclim.py
arbennett: I'll add this.  Just a note, I have it set up so that numba doesn't allow object mode, and will fail at compile time.

metsim/methods/mtclim.py
arbennett: `solar_geom` unfortunately can't be turned into a Pandas dataframe due to the structure of `tiny_rad_fract`.  Since that's the case we don't have anything to call `.where` on.

metsim/disaggregate.py
arbennett: The reason this is using scipy is precisely because I don't want to recreate the wheel. `(time, temp)` are a set of coordinates to interpolate over, not Series.  To turn the interpolation step into 1 line we would have to get these into series, which to my knowledge is going to take more than 2 lines.  I just played around with it for a bit and don't really know how to make this happen without adding way more logic than is necessary. Maybe you could figure something out to make this work, but you'll have to rewrite the `set_min_max_hour` function to do this. 

As it stands, this works, and ends up being more efficient than a replacement using pandas exclusively, since we only have to keep 2 points per day instead of all of them.

metsim/constants.py
```diff
@@ -0,0 +1,56 @@
...
+DAYS_PER_YEAR = 365.25
+SEC_PER_DAY   = 86400
+MIN_PER_HOUR  = 60
+HOURS_PER_DAY  = 24
+SEC_PER_HOUR  = 3600
+STEFAN_B = 5.669e-8     # (W m^-22 K^-4) Stefan Boltzmann constant 
+TDAYCOEF= 0.45          # (dim) daylight air temperature coefficient
+
+BAR_TO_MBAR = 1./1000
+CM_TO_MM = 1./10
```
bartnijssen: Confusing ... since 1 bar equals 1000 mbar and 1 cm equals 10 mm, these seem to be exactly opposite to what their name indicates (assuming this is a multiplication, not a division). At the very least provide a use case in a comment.

metsim/constants.py
```diff
@@ -0,0 +1,56 @@
+"""
+Stores default constants, parameters, and options.  
+"""
```
bartnijssen: Would be good to organize the file so that physical constants (value is always the same), conversion parameters (`SECPERRAD`) parameters (tunable) and options are organized in blocks with clear identification what is what.

metsim/constants.py
```diff
@@ -0,0 +1,56 @@
...
+                       #    Note:  Make sure that 3600 % SRADDT == 0
+MA= 28.9644e-3         # (kg mol-1) molecular weight of air
+R= 8.3143              # (m3 Pa mol-1 K-1) gas law constant
+G_STD= 9.80665         # (m s-2) standard gravitational accel.
+P_STD= 101325.0        # (Pa) standard pressure at 0. m elevation
+T_STD= 288.15          # (K) standard temp at 0. m elevation
+CP= 1010.0             # (J kg-1 K-1) specific heat of air
+LR_STD= 0.0065         # (-K m-1) standard temperature lapse rate
+ 
+KELVIN= 273.15
```
bartnijssen: Include a comment line for each value that includes units

metsim/constants.py
bartnijssen: Note that `CMPERMM` would be more consistent with the earlier (`SECPERRAD` or `SEC_PER_DAY`). Preferably standardize on underscore usage as well, but I understand that is likely inherited.

metsim/constants.py
```diff
@@ -0,0 +1,56 @@
...
+STEFAN_B = 5.669e-8     # (W m^-22 K^-4) Stefan Boltzmann constant 
+TDAYCOEF= 0.45          # (dim) daylight air temperature coefficient
+
+BAR_TO_MBAR = 1./1000
+CM_TO_MM = 1./10
+
+
+# parameters for the snowpack algorithm
+SNOW_TCRIT= -6.0   # (deg C) critical temperature for snowmelt
+SNOW_TRATE= 0.042  # (cm/degC/day) snowmelt rate
```
bartnijssen: Make it clear that these parameters come from the old mtclim code

metsim/disaggregate.py
```diff
@@ -7,107 +7,172 @@
...
     return df_disagg.fillna(method='ffill')
 
 
-def set_min_max_hour(disagg_rad, n_days):
+def set_min_max_hour(disagg_rad, n_days, ts):
     """
     TODO
     """
-    ts = float(params['time_step'])
     rad_mask = 1*(disagg_rad > 0)
```
bartnijssen: Why the `1*`? Just for my own edification.

metsim/disaggregate.py
```diff
@@ -7,107 +7,172 @@
...
     diff_mask = np.diff(rad_mask)
-    rise_times = np.where(diff_mask>0)[0] * ts 
+    rise_times = np.where(diff_mask>0)[0] * ts
     set_times = np.where(diff_mask<0)[0] * ts
-    t_Tmax = 0.67 * (set_times - rise_times) + rise_times
-    t_Tmin = rise_times - float(params['time_step'])
-    return t_Tmin, t_Tmax
+    t_t_max = 0.67 * (set_times - rise_times) + rise_times
```
bartnijssen: Pull out magic number (`0.67`) as much as possible and replace them by a variable. In this case, the value would be added to `constants.py` with a brief description

metsim/disaggregate.py
bartnijssen: Open an issue to evaluate the use of pandas here, but leave as is for now. I do want you to add a comment to the code, that explains why the `scipy.interpolate.PchipInterpolator()` method was selected.

metsim/disaggregate.py
```diff
@@ -7,107 +7,172 @@
...
     """
-    return longwave.resample(params['time_step']+'T').fillna(method='ffill')
+    return wind.resample('{:0.0f}T'.format(ts)).fillna(method='ffill')
 
 
-def wind(wind):
+def relative_humidity(vapor_pressure, temp):
     """
-    Wind is assumed constant throughout the day
+    TODO
```
bartnijssen: Dont't leave `TODO`s in the code. It's fine for this PR, but let's make sure we update it before we release 1.0.0

metsim/disaggregate.py
```diff
@@ -7,107 +7,172 @@
...
     """
-    return wind.resample(params['time_step']+'T').fillna(method='ffill')
-
-
-def shortwave(sw_rad, daylength, day_of_year, tiny_rad_fract):
+    rh = 100 * (vapor_pressure / cnst.BAR_TO_MBAR) / svp(temp)
+    return rh.where(rh < 100, 100.0)
```
bartnijssen: replace `100` by `RH_UPPER_BOUND` or something like it

metsim/disaggregate.py
```diff
@@ -7,107 +7,172 @@
...
-
-
-def shortwave(sw_rad, daylength, day_of_year, tiny_rad_fract):
+    rh = 100 * (vapor_pressure / cnst.BAR_TO_MBAR) / svp(temp)
+    return rh.where(rh < 100, 100.0) 
+
+
+def vapor_pressure(hum_daily, temp, t_Tmin, n_out, ts):
+    """Calculate vapor pressure"""
+    # Scale down to milibar
```
bartnijssen: `millibar`

metsim/disaggregate.py
```diff
@@ -7,107 +7,172 @@
...
+    vp_daily = hum_daily * cnst.BAR_TO_MBAR 
+    # Linearly interpolate the values
+    interp = scipy.interpolate.interp1d(t_Tmin, vp_daily, fill_value='extrapolate')
```
bartnijssen: Comment on choice of `interpolate` method

metsim/disaggregate.py
```diff
@@ -7,107 +7,172 @@
...
+        'DEFAULT'    : lambda vp : vp,
+        'TVA'        : lambda vp : 0.74 + 0.0049 * vp,
+        'ANDERSON'   : lambda vp : 0.68 + 0.036 * np.power(vp, 0.5),
+        'BRUTSAERT'  : lambda vp : 1.24 * np.power(vp/air_temp, 0.14285714),
+        'SATTERLUND' : lambda vp : 1.08 * (
+            1 - np.exp(-1 * np.power(vp, (air_temp/2016)))),
+        'IDSO'       : lambda vp : 0.7 + 5.95e-5 * vp * np.exp(1500/air_temp),
+        'PRATA'      : lambda vp : (1 - (1 + (46.5*vp/air_temp)) *
+            np.exp(-np.sqrt((1.2 + 3. * (46.5*vp/air_temp)))))
+        }
```
bartnijssen: Nice

metsim/disaggregate.py
```diff
@@ -7,107 +7,172 @@
...
+                   / (cnst.HOURS_PER_DAY/360)))
+
     # Tinystep represents a daily set of values - but is constant across days
-    tinystep= np.arange(consts['HOURS_PER_DAY'] * tiny_step_per_hour) - tiny_offset
-    tinystep[np.array(tinystep<0)] += consts['HOURS_PER_DAY'] * tiny_step_per_hour
-    tinystep[np.array(tinystep>(24*tiny_step_per_hour-1))] -= consts['HOURS_PER_DAY'] * tiny_step_per_hour
-    chunk_sum = lambda x : np.sum(x.reshape((len(x)/120, 120)), axis=1)
+    tinystep = np.arange(cnst.HOURS_PER_DAY * tiny_step_per_hour) - tiny_offset
+    tinystep[np.array(tinystep<0)] += cnst.HOURS_PER_DAY * tiny_step_per_hour
+    tinystep[np.array(tinystep>(24*tiny_step_per_hour-1))] -= (
```
bartnijssen: replace 24 with `cnst.HOURS_PER_DAY`

metsim/disaggregate.py
```diff
@@ -7,107 +7,172 @@
...
+    tmp_rad = sw_rad * daylength / cnst.SEC_PER_HOUR
     n_days = len(tmp_rad)
-    ts_per_day = consts['HOURS_PER_DAY'] * (consts['MIN_PER_HOUR']/int(params['time_step']))
-    disaggrad = np.zeros(n_days*ts_per_day + 1)
-    tiny_offset = (params.get("theta_l", 0) - params.get("theta_s", 0) / (consts['HOURS_PER_DAY']/360))
-   
+    ts_per_day = cnst.HOURS_PER_DAY * (cnst.MIN_PER_HOUR/int(params['time_step']))
+    disaggrad = np.zeros(int(n_days*ts_per_day) + 1)
+    tiny_offset = ((params.get("theta_l", 0) - params.get("theta_s", 0)
+                   / (cnst.HOURS_PER_DAY/360)))
```
bartnijssen: replace `360`

metsim/methods/mtclim.py
```diff
@@ -25,75 +20,65 @@ def run(forcing: pd.DataFrame, disagg=True):
...
 
 
-def calc_t_air(df: pd.DataFrame):
+def calc_t_air(df: pd.DataFrame, params: dict):
     """ 
     Adjust temperatures according to lapse rates 
     and calculate t_day
     """
-    dZ = (params['site_elev'] - params['base_elev'])/1000.
+    dZ = (df['elev'][0] - params['base_elev'])/1000.
```
bartnijssen: replace `1000.`

metsim/methods/mtclim.py
```diff
@@ -106,127 +91,90 @@ def _calc_tfmax(precip, dtr, sm_dtr):
...
+    pa = atm_pres(df['elev'][0])
     yday = df['day_of_year'] - 1 
     df['dayl'] = sg['daylength'][yday]
-
+ 
+    # Calculation of tdew and swrad. tdew is iterated on until
+    # it converges sufficiently 
+    tdew_old = tdew
+    tdew, pva = sw_hum_iter(df, sg, pa, pva, parray, dtr)
+    while(np.sqrt(np.mean((tdew-tdew_old)**2)) > 1e-3):
```
bartnijssen: replace `1e-3` with something like `TDEW_CONVERGENCE_THRESHOLD`

metsim/methods/mtclim.py
```diff
@@ -106,127 +91,90 @@ def _calc_tfmax(precip, dtr, sm_dtr):
...
-            'CLOUD_DEARDORFF' : lambda x : tskc * 1.0 + (1-tskc) * x
-            }
-    tskc = df['tskc']
-    air_temp = df['t_day'] + consts['KELVIN'] 
-    emissivity_clear = emissivity_calc[options['LW_TYPE'].upper()](air_temp)
-    emissivity = cloud_calc[options['LW_CLOUD'].upper()](emissivity_clear) 
-    df['lwrad'] = emissivity * consts['STEFAN_B'] * np.power(air_temp, 4)
+    df['pet'] = parray 
+    tmink = df['t_min'] + cnst.KELVIN
+    tdew = tmink*(-0.127 + 1.121*(1.003 - 1.444*ratio + 12.312*np.power(ratio, 2)
```
bartnijssen: I think here it's OK to maintain the literal values. Would be good to state where the function comes from.

metsim/metsim.py
```diff
@@ -9,41 +9,67 @@
...
+        "domain":'',
+        "out_dir":'',
+        "start":'',
+        "stop":'',
+        "time_step":'',
+        "out_format":'',
+        "site_elev": 0.,
+        "base_elev": 0.,
+        "t_max_lr": 0.0065,
+        "t_min_lr": 0.0065,
```
bartnijssen: Make `t_max_lr` and `t_min_lr` required configuration parameters.

metsim/metsim.py
```diff
@@ -55,87 +81,101 @@ def run(self, method, forcings):
...
                 'ascii'  : self.write_ascii
                 }
-        dispatch[self.params.get('out_format', 'netcdf')](data)
-         
+        dispatch[MetSim.params.get('out_format', 'netcdf').lower()](data)
+
 
     def init_netcdf(self):
         """Initialize the output file"""
         print("Initializing netcdf...")
```
bartnijssen: remove debugging statements (or wrap them in some `verbose` option or logging function)

metsim/metsim.py
```diff
@@ -55,87 +81,101 @@ def run(self, method, forcings):
...
-        out_file = os.path.join(self.params['out_dir'], "forcing.nc")
+        out_dir = MetSim.params.get('out_dir', './results/')
```
bartnijssen: don't hardwire directories

metsim/metsim.py
```diff
@@ -55,87 +81,101 @@ def run(self, method, forcings):
...
         # Create output variables
-        for varname in self.params['out_vars']:
+        for varname in MetSim.params['out_vars']:
             self.output.createVariable(varname, 'd', ('time', 'lat','lon'))
 
 
     def write_netcdf(self, data:dict):
-        """Write out to the output file"""
+        """Write out as NetCDF to the output file"""
         print("Writing netcdf...")
```
bartnijssen: Remove debugging statements

metsim/metsim.py
```diff
@@ -55,87 +81,101 @@ def run(self, method, forcings):
...
+            for lon in data[lat].keys():
+                for varname in MetSim.params['out_vars']:
+                    self.output.variables[varname][:, lat, lon] = (
+                            data[lat][lon][varname].values)
 
 
     def write_ascii(self, data: dict):
-        """ TODO """
+        """Write out as ASCII to the output file"""
         print("Writing ascii...")
```
bartnijssen: Remove debugging statements (or wrap them)

metsim/metsim.py
```diff
@@ -55,87 +81,101 @@ def run(self, method, forcings):
...
         n_days = len(dates)
         precip = [] # Short unsigned int
         t_max  = [] # Short int
         t_min  = [] # Short int
         wind   = [] # Short int
-        
+
         # Pack these for nicer syntax in the loop
         var_name = [precip, t_max, t_min, wind]
         scale = [40.0, 100.0, 100.0, 100.0]
```
bartnijssen: scale values should be read from the configuration file (or the netcdf file in case of netcdf)

metsim/metsim.py
```diff
@@ -191,25 +231,25 @@ def read_netcdf(self, fpath:str) -> xr.Dataset:
...
+
+
     def read(self, fpath:str) -> xr.Dataset:
         """
-        Dispatch to the right function based on the file extension 
+        Dispatch to the right function based on the file extension
         """
         ext_to_fun = {
                 '.bin'   : self.read_binary,
                 '.nc'    : self.read_netcdf,
```
bartnijssen: What is the default netcdf version for `nc`? Provide a comment on each line

bartnijssen:
Running it through a linter would be great as well. It's nice if our code is generally PEP8 compliant.

metsim/metsim.py
arbennett: That's not hardwired, but the default if none is given.

metsim/constants.py
arbennett: Documentation is forthcoming, as mentioned in the PR description.

metsim/disaggregate.py
arbennett: In this case, what would be a good name?

metsim/constants.py
arbennett: I'm not sure which things should be considered tunable, but they should all go in the MetSim parameters dictionary instead of the constants module, so they can be specified in the configuration file.

metsim/constants.py
arbennett: Yeah, these should probably be changed to `MBAR_PER_BAR` and `MM_PER_CM` and flipped for clarity.

metsim/disaggregate.py
arbennett: This coerces the booleans to integers, so they can be `np.diff`'d.

metsim/disaggregate.py
arbennett: This will be addressed in a pull request that handles fixing of documentation.

metsim/methods/mtclim.py
arbennett: This will be addressed with added documentation.

metsim/metsim.py
arbennett: Verbose mode will be added in a future pull request.

metsim/metsim.py
arbennett: Adding verbose mode is part of a future pull request.

metsim/metsim.py
arbennett: Adding verbose mode is part of a future pull request.

metsim/metsim.py
arbennett: I suppose the default is NetCDF 4, since we use the `netCDF4` library.  It can handle version 3 as well, if I recall correctly.

metsim/metsim.py
arbennett: Added an issue for this.

metsim/disaggregate.py
arbennett: Done, and the documentation will have an explanation of why that method was chosen.

metsim/constants.py
bartnijssen: OK - opening an issue is fine - all I want here is separating true physical constants (stefan-boltmann, etc.) from the others

metsim/disaggregate.py
bartnijssen: Please fix

metsim/metsim.py
bartnijssen: Throw an error if none is given rather than assume one.

metsim/metsim.py
bartnijssen: So what's the difference between `nc` and `nc4` in that case?

bartnijssen:
@jhamman : go ahead and approve your review for this PR. I can override it, but don't want to. @arbennett has opened issues for the pandas aspect.

@arbennett : A few final questions from me. After that I'll merge

metsim/metsim.py
arbennett: No difference in functionality, but both are commonly used as file extensions for NetCDF files, so we want them to both go to the same function.

bartnijssen:
I opened an issue (issue #20) to address the code formatting. We don't need to hold up this PR for it, but we should do it before milestone 1.0.0

PR #30, opened 2017-02-01
Adding parallelism

arbennett:
This pull request gets parallelism working.  Parallelism is implemented by splitting up the domain of the problem into `n` chunks, where `n` can be specified as a command line option.  A dedicated IO process is launched, and then `n` worker processes are set in motion.  The IO is handled with a simple shared memory queue.

To get this working it was required to add some preprocessing methods which examine the given input to determine the extent on which to parallelize.  For simplicity, if given binary input, they will first be translated into an Xarray Dataset, which is how the NetCDF input is represented.


metsim/methods/mtclim.py
```diff
@@ -89,72 +88,72 @@ def _calc_tfmax(precip, dtr, sm_dtr):
...
     t_final = t_tmax * df['tfmax']
 
-    # Snowpack contribution 
+    # Snowpack contribution
     sc = np.zeros_like(df['swe'])
     if (cnst.MTCLIM_SWE_CORR):
         inds = np.logical_and(df['swe'] > 0.,  daylength[yday] > 0.)
         sc[inds] = (1.32 + 0.096 * df['swe'][inds]) * 1.0e6 / daylength[yday][inds]
-        sc = np.maximum(sc, 100.)  # JJH - this is fishy 
+        sc = np.maximum(sc, 100.)  # JJH - this is fishy
```
hhelgason: Just a reminder to address the fishiness.

metsim/metsim.py
```diff
@@ -42,71 +47,143 @@ def __init__(self, params:dict):
...
         """
         # Where we will store the results for writing out
         out_dict = {}
         method = MetSim.methods[MetSim.params.get('method', 'mtclim')]
 
-        # Coerce the forcings into a list if it's not one
-        if type(forcings) is not list:
-            forcings = [forcings]
-
         # Do the forcing generation and dissaggregation if required
```
hhelgason: disaggregation

metsim/disaggregate.py
```diff
@@ -17,19 +17,19 @@ def disaggregate(df_daily, params, solar_geom):
     stop = params['stop'] + pd.Timedelta('1 days')
     dates_disagg = pd.date_range(params['start'], stop, freq=params['time_step']+'T')
     df_disagg = pd.DataFrame(index=dates_disagg)
-    n_days = len(df_daily['day_of_year'])
+    n_days = len(df_daily.index)
```
jhamman: This can just be `len(df_daily)`

metsim/disaggregate.py
```diff
@@ -75,17 +75,21 @@ def temp(df_daily, df_disagg, t_t_min, t_t_max, ts):
...
 
 
-def precip(precip, ts):
+def prec(prec, ts):
     """
     Splits the daily precipitation evenly throughout the day
     """
     scale = int(ts) / (cnst.MIN_PER_HOUR * cnst.HOURS_PER_DAY)
-    return (precip*scale).resample('{:0.0f}T'.format(ts)).fillna(method='ffill')
+    return (prec*scale).resample('{:0.0f}T'.format(ts)).fillna(method='ffill')
```
jhamman: spaces around `*`

metsim/disaggregate.py
```diff
@@ -75,17 +75,21 @@ def temp(df_daily, df_disagg, t_t_min, t_t_max, ts):
...
 
     # Interpolate the values
-    interp = scipy.interpolate.PchipInterpolator(time, temp, extrapolate=True)
-    temps = interp(ts * np.arange(0, len(df_disagg.index)))
+    try:
+        interp = scipy.interpolate.PchipInterpolator(time, temp, extrapolate=True)
+        temps = interp(ts * np.arange(0, len(df_disagg.index)))
+    except ValueError:
+        temps = np.full(len(df_disagg.index), np.nan)
+
```
jhamman: I'm coming back to this. I think @orianac asked the same question. Why do we need the try/except statement?  Also you don't need the `.index` on the `len(df_disagg)` occurrences.

metsim/disaggregate.py
```diff
@@ -99,20 +103,22 @@ def relative_humidity(vapor_pressure, temp):
     """
     TODO
     """
-    rh = 100 * (vapor_pressure / cnst.MBAR_PER_BAR) / svp(temp)
-    return rh.where(rh < cnst.MAX_PERCENT, cnst.MAX_PERCENT) 
+    rh = cnst.MAX_PERCENT * cnst.MBAR_PER_BAR * (vapor_pressure/svp(temp))
```
jhamman: spaces around `/`

metsim/disaggregate.py
```diff
@@ -99,20 +103,22 @@ def relative_humidity(vapor_pressure, temp):
...
-    vp_disagg = interp(ts * np.arange(0, n_out))
-
-     # Account for situations where vapor pressure is higher than saturation point
-    vp_sat = svp(temp) * cnst.MBAR_PER_BAR
+    try:
+        interp = scipy.interpolate.interp1d(t_Tmin, vp_daily/cnst.MBAR_PER_BAR,
+                                            fill_value='extrapolate')
+        vp_disagg = interp(ts * np.arange(0, n_out))
+    except ValueError:
+        vp_disagg = np.full(len(temp), np.nan)
```
jhamman: see comments above on PEP8 and error handling.

metsim/methods/mtclim.py
```diff
@@ -54,30 +54,29 @@ def calc_t_air(df: pd.DataFrame, params: dict):
     df['t_day'] = ((t_max - t_mean) * cnst.TDAY_COEF) + t_mean
 
 
-def calc_precip(df: pd.DataFrame, params: dict):
-    """ Adjust precipitation according to isoh """
-    df['precip'] = (df['precip'] * (df.get('site_isoh', 1) / df.get('base_isoh', 1)))
+def calc_prec(df: pd.DataFrame, params: dict):
+    """ Adjust precitation according to isoh """
+    df['prec'] = (df['prec'] * (df.get('site_isoh', 1) / df.get('base_isoh', 1)))
```
jhamman: perhaps better written as:

```Python
df['prec'] *= (df.get('site_isoh', 1) / df.get('base_isoh', 1))
```

metsim/metsim.py
```diff
@@ -42,71 +47,143 @@ def __init__(self, params:dict):
...
-                                                .to_dataframe(), MetSim.params,
-                                                disagg=True)
-            # Write out
-            self.write(out_dict)
-
-        if MetSim.params.get('out_format', 'netcdf') == 'netcdf':
-            self.output.close()
+        # Do the forcing generation and disaggregation if required
+        for i, j in locations:
+            print("Processing {} {}".format(i, j))
```
jhamman: Use logging module.

metsim/disaggregate.py
arbennett: If an input file is given, say for CONUS in a format which is just a bounding box, then a lot of the ocean gridcells won't have associated data, in which case the interpolation would fail.

metsim/metsim.py
arbennett: Verbose mode is going to be part of my documentation and cleanup sweep coming up next, this will be taken care of then.

metsim/disaggregate.py
jhamman: I think this is too deep in the code to be handling domain/mask issues.  Ideally, we just iterate over the cells that we care about.

metsim/metsim.py
bartnijssen: OK - but use the logging module to implement verbose mode. It's pretty easy to use and will allow you to configure different levels of verbosity.

metsim/disaggregate.py
bartnijssen: I am confused by this anyway - why would we even be handling ocean cells if they don't have associated data. I think I am basically commenting on the same thing as @jhamman . We should only run over the 'active' grid cells, so the interpolator should never be called on ocean grid cells (or cells outside the active mask). This is important for the parallelism as well, because you don't want large parts that are outside the mask be assigned to a process that then just sits idle.

metsim/disaggregate.py
```diff
@@ -75,17 +75,21 @@ def temp(df_daily, df_disagg, t_t_min, t_t_max, ts):
...
 
 
-def precip(precip, ts):
+def prec(prec, ts):
     """
     Splits the daily precipitation evenly throughout the day
     """
     scale = int(ts) / (cnst.MIN_PER_HOUR * cnst.HOURS_PER_DAY)
-    return (precip*scale).resample('{:0.0f}T'.format(ts)).fillna(method='ffill')
+    return (prec * scale).resample('{:0.0f}T'.format(ts)).fillna(method='ffill')
```
bartnijssen: Ideally you would split your PRs into more specific elements - these changes are not associate with "adding parallelism", so they're like unrelated things being tacked onto a bill in congress. The reason it matters is that it is easier to roll back specific PRs (in case something is wrong) if they are more atomic. Leave as-is for now

metsim/methods/mtclim.py
```diff
@@ -89,72 +88,72 @@ def _calc_tfmax(precip, dtr, sm_dtr):
...
-    # Calculate annual total precip
-    sum_precip = df['precip'].values.sum()
-    ann_precip = (sum_precip / params['n_days']) * cnst.DAYS_PER_YEAR
-    if ann_precip == 0.0:
-        ann_precip = 1.0
+    # Calculate annual total prec
+    sum_prec = df['prec'].values.sum()
+    ann_prec = (sum_prec / params['n_days']) * cnst.DAYS_PER_YEAR
+    if ann_prec == 0.0:
+        ann_prec = 1.0
```
bartnijssen: Requires a comment as to why you are resetting `ann_prec`

metsim/methods/mtclim.py
```diff
@@ -89,72 +88,72 @@ def _calc_tfmax(precip, dtr, sm_dtr):
...
-            window[:90] = df['precip'][:90]
+            window[:90] = df['prec'][:90]
 
-        parray = (window.rolling(window=90, win_type=win_type, axis=0) 
+        parray = (window.rolling(window=90, win_type=win_type, axis=0)
                     .mean()[90:] * cnst.DAYS_PER_YEAR)
 
-    # Convert to mm 
+    # Convert to cm
     parray = parray.where(parray>80.0, 80.0) / cnst.MM_PER_CM
```
bartnijssen: I realize that this is not changed in this PR, but what's special about `80`? Please add comment.

metsim/methods/mtclim.py
bartnijssen: I think `100.` needs to be replaced by `cnst.MAX_PERCENT`

metsim/metsim.py
```diff
@@ -42,71 +47,143 @@ def __init__(self, params:dict):
...
+                         "netcdf" : self.netdf_in_preprocess}
+        met_data = in_preprocess[MetSim.params['in_format']](job_list)
+
+        # Start up the IO Process
+        io_process = Process(target=self.create_io_process, args=(self.queue,))
+        io_process.start()
+
+        # Split the jobs up into groups based on the number
+        # of desired processes
+        self.locations = np.array_split(np.array(self.locations),  MetSim.params['nprocs'])
```
bartnijssen: Filter the locations for active grid cells before farming them out. This should get rid of the need for the `try: ... except: ...` statements around the interpolator deeper in the code and it makes sure each process has the same number of *active* cells

metsim/metsim.py
```diff
@@ -42,71 +47,143 @@ def __init__(self, params:dict):
...
+                               MetSim.params, disagg=True))
+        queue.put(out_dict)
 
 
     def find_elevation(self, lat: float, lon: float) -> float:
         """ Use the domain file to get the elevation """
-        lat_idx = np.abs(self.lats - lat).argmin()
-        lon_idx = np.abs(self.lons - lon).argmin()
-        return self.elev[lat_idx, lon_idx]
+        lat_idx = np.abs(self.domain_lats - lat).argmin()
```
bartnijssen: Is this robust? Would be good to see some tests for this.

metsim/metsim.py
```diff
@@ -42,71 +47,143 @@ def __init__(self, params:dict):
...
+        in_forcing = Dataset(filename, 'r')
+        self.locations = list(itertools.product(in_forcing['lat'], in_forcing['lon']))
+        self.lats = np.array(in_forcing['lat'])
+        self.lons = np.array(in_forcing['lon'])
+        self.lat_idx = {str(lat): i for i, lat in enumerate(self.lats)}
+        self.lon_idx = {str(lon): j for j, lon in enumerate(self.lons)}
+        in_forcing.close()
+        # Now read in the file properly
+        met_data = self.read(filename)
+        return met_data
```
bartnijssen: Modify this to only return the active grid cells. Keep the extent to write out empty grid cells as `_FillValue`

metsim/metsim.py
```diff
@@ -42,71 +47,143 @@ def __init__(self, params:dict):
...
+        # Binary forcing files have naming format $NAME_$LAT_$LON
+        sets = [os.path.basename(fpath).split("_") for fpath in job_list]
+        self.locations = [(float(s[1]), float(s[2])) for s in sets]
+        self.lats = np.unique(sorted([float(s[1]) for s in sets]))
+        self.lons = np.unique(sorted([float(s[2]) for s in sets]))
+        self.lat_idx = {str(lat): i for i, lat in enumerate(self.lats)}
+        self.lon_idx = {str(lon): j for j, lon in enumerate(self.lons)}
+
+        # Creates the master dataset which will be used to parallelize
+        met_data = xr.Dataset(coords={'time' : MetSim.params['dates'],
```
bartnijssen: No need to have a master dataset that covers the entire domain. You only need to keep track of active cells and process those. Then inactive ones you can handle as part of the write statement only. This will speed up the process and reduce memory requirements.

metsim/methods/mtclim.py
arbennett: 80mm was the minimum annual precipitation set in the VIC code, to avoid very large PET to annual precip ratios.  I don't know why 80 was specifically chosen.

metsim/disaggregate.py
arbennett: @jhamman is working on this now.  I'll take out the `try...except` blocks here.

metsim/metsim.py
arbennett: @jhamman and I have been discussing better ways to handle the parallelism and writeout process and will be addressed in a future pull request.

metsim/methods/mtclim.py
arbennett: Looks like we can actually get rid of/simplify that section of code entirely.  I'll make an issue about it and address it in a future pull request.  I think it's better to do that way to simplify merging, since @jhamman has done a bunch of cleanup in a separate pull request (where this should belong).

jhamman:
@bartnijssen - I think we can merge this now. @arbennett and I have discussed a followup/refactor PR that I am currently working to speed up write I/O which is currently are pretty big bottleneck. I'm going to merge if no one has any objections.

bartnijssen:
@jhamman - go ahead and merge. I tried to do it on my phone, but apparently "pull requests that have a failing status can't be merged on a phone"

PR #31, opened 2017-02-09
Test

arbennett:
Adding a preliminary set of tests.  They check for operability of input/output/method combinations, setup, mtclim, disaggregation, and runs numerical checks on a dataset that's been validated against VIC4.2's disaggregation.


bartnijssen:
This one is a bit hard to review till #30 is done, so I will wait

arbennett:
I'm going to close this pull request because it would be overtly burdensome to rework this into the framework that's now being developed.  After my next pull request fixing a bunch of stuff with parallelism this will be reintroduced.

PR #32, opened 2017-02-10
Cleanup

jhamman:
cleanup following #30.


jhamman:
@arbennett - any issue with this PR going in?

PR #34, opened 2017-02-13
Feature/daily output

jhamman:
cc @arbennett 

will add more discussion later...


examples/example.conf
```diff
@@ -11,33 +11,24 @@ start = 1949/1/1:0
...
-               "longwave", 
-               "wind"
-              ]
+# Variables given
+in_vars = [
+           "prec", 
+           "t_max", 
+           "t_min", 
+           "wind"
+          ]
```
jhamman: Update with new example...

examples/example_nc.conf
```diff
@@ -14,19 +14,19 @@ stop = 1950/1/31
...
 method = mtclim
 
 # Variables given
 in_vars = [
-           "prec", 
-           "t_max", 
-           "t_min", 
+           "prec",
+           "t_max",
+           "t_min",
```
jhamman: Update with new example...

jhamman:
since @arbennett's #35 is the follow up to these changes, I will merge now.

Issue #23, opened 2017-01-30
Add configuration validation function


arbennett:
We should validate that all of the necessary parameters are set, and that they have valid values.

arbennett:
Fixed in #37

jhamman:
I'm going to reopen this.  I just specified a invalid option and MetSim only gave me this error message:

```Invalid configuration given.  Use `ms -h` for more information.```

running `ms-h` gives:

```$ ms -h
usage: ms [-h] [-n N_PROCESSES] config

positional arguments:
  config

optional arguments:
  -h, --help            show this help message and exit
  -n N_PROCESSES, --n-processes N_PROCESSES```

Which doesn't help me figure out which configuration option caused the error. We basically need to do a better job explaining why we exit.

arbennett:
Could you show what your parameter set was?

arbennett:
This was reopened with a problem that this issue wasn't trying to address. Reclosing.

Issue #20, opened 2017-01-30
Separate parameters from constants


arbennett:
The `constants` module may have some variables that we should allow users to tune themselves via the configuration file.  @bartnijssen @jhamman do either of you have opinions where we draw those lines?

arbennett:
Looking through it seems like the following should be moved:

* `SW_RAD_DT`
* `TDAY_COEF`
* `SW_PREC_THRESH`
* `MTCLIM_SWE_CORR`
* `LW_CLOUD`
* `LW_TYPE`

arbennett:
This was solved in #37.

PR #36, opened 2017-02-17
Move parameters and add some safety checks

arbennett:
This pull request addresses #20 and #23.


metsim/constants.py
```diff
@@ -25,6 +25,7 @@
 SEC_PER_DAY = 86400
 MIN_PER_HOUR = 60
 HOURS_PER_DAY = 24
+MIN_PER_DAY = MIN_PER_HOUR * HOURS_PER_DAY
```
bartnijssen: If you are calculating `MIN_PER_DAY` this way, then you might as well extend the same treatment to `SEC_PER_DAY`

metsim/metsim.py
```diff
@@ -25,7 +25,7 @@
 now = tm.ctime(tm.time())
 user = getuser()
 
-attrs = {'pet': {'units': 'mm d-1', 'long_name': 'potential evaporation',
+attrs = {'pet': {'units': 'cm d-1', 'long_name': 'potential evaporation',
```
bartnijssen: That's a weird unit. Please make this `mm d-1` and do the conversion as required.

arbennett:
I just noticed something went horribly wrong with the commit history in this pull request.  I'm going to close this and clean that up - a new pull request will be created so that we don't have 2x the commits as actually happened.

PR #58, opened 2017-03-07
Refactor/parallel update

jhamman:
This PR does away with the domain decomposition using `CHUNK_SIZE`. While I liked where we were headed with the previous approach, it had a fatal flaw, it passed the full forcing dataset across processes and for large datasets, this crashed (`PickleError`). So what I've done here is sacrifice some efficiency for operability.  I'll open an issue to revisit this subject.


bartnijssen:
It breaks in the tests on unpacking the results with which the run is compared (I think). This refactor should not have affected the actual output. I suspect that this will be a pretty clean fix (but don't really know where it is.

arbennett:
Before I say I'm okay merging this in I have one question.  Wouldn't it be possible to keep the chunking in, but uses slices to sub-select out from the input forcing data?

jhamman:
Yes, that is possible. It won't save us any time though because the point-wise indexing in xarray is currently a pure-Python loop.

PR #65, opened 2017-03-16
Bug/non hourly disaggregation

### LINKED ISSUE #12: Timesteps of greater than 60 minutes causes error on shortwave dissaggregation
This is caused by `tinystep` and `chunk_sum`.

Error message is:

`
````````Traceback (most recent call last):
  File "./ms", line 83, in <module>
    main()
  File "./ms", line 78, in main
    ms.run(setup['method'], setup['forcings'])
  File "/home/bzq/research/MetSim/metsim/metsim.py", line 63, in run
    .to_dataframe(), disagg=True)
  File "/home/bzq/research/MetSim/metsim/methods/mtclim.py", line 37, in run
    forcing = disaggregate(forcing, sg)
  File "/home/bzq/research/MetSim/metsim/disaggregate.py", line 23, in disaggregate
    solar_geom['tiny_rad_fract']))
  File "/home/bzq/research/MetSim/metsim/disaggregate.py", line 106, in shortwave
    chunk_sum(rad[list(tinystep)]) * tmp_rad[day])
ValueError: could not broadcast input array from shape (24) into shape (12)`

### PR DESCRIPTION
**Pull request by arbennett**
- [x] closes #12
 - [x] tests added / passed

This fixes the shortwave radiation bug where disaggregation of non-hourly time lengths caused failures.  The trouble was caused by the chunk of `tiny_rad_fract` being aggregated was incorrect.

### CODE REVIEW

File: metsim/disaggregate.py [hunk 1]
```diff
@@ -363,16 +363,20 @@ def shortwave(sw_rad: pd.Series, daylength: pd.Series, day_of_year: pd.Series,
     tinystep[inds] += cnst.HOURS_PER_DAY * tiny_step_per_hour
     inds = np.array(tinystep > (cnst.HOURS_PER_DAY * tiny_step_per_hour-1))
     tinystep[inds] -= (cnst.HOURS_PER_DAY * tiny_step_per_hour)
+    tinystep = np.array(tinystep).astype(np.int32)
```
jhamman: can you use `np.asarray()` instead here (and above). You can also specify the dtype there too (i.e. `np.asarray(tinystep, dtype=np.in32)`

metsim/disaggregate.py
```diff
@@ -363,16 +363,20 @@ def shortwave(sw_rad: pd.Series, daylength: pd.Series, day_of_year: pd.Series,
...
+        return np.sum(x.reshape((int(len(x)/chunk_size), chunk_size)), axis=1)
 
     for day in range(n_days):
         rad = tiny_rad_fract[day_of_year[day] - 1]
         dslice = slice(int(day * ts_per_day), int((day + 1) * ts_per_day))
-        disaggrad[dslice] = (
-            chunk_sum(rad[np.array(tinystep).astype(np.int32)]) * tmp_rad[day])
-
+        rad_chunk = rad[np.array(tinystep).astype(np.int32)]
+        disaggrad[dslice] = chunk_sum(rad_chunk) * tmp_rad[day]
```
jhamman: this for loop should be easy to numba-fy, yes?

arbennett: Unfortunately not in its current form.  I have played with it in the past, and generally get a bit of a slowdown if we jit the whole `shortwave` function.  I think if we moved it to a more "pure" python implementation we could probably get some speedup with numba, but I have no idea how much.  I'm much less optimistic about this now, since we aren't able to cache the compiled functions.

jhamman: Okay. No worries. Let's revisit this once we have a v.1.0 release.

Issue #21, opened 2017-01-30
Pull out annual aggregates into the domain file


arbennett:
Some aggregates such as effective annual precipitation are currently calculated within the MTCLIM implementation.  Having them be estimated using the given forcing makes these values vary from run to run, depending on the length of the input timeseries.  This can lead to inconsistencies.

For example, if 2 years of data are supposed to be disaggregated the results will be different if each year is processed separately than if they are processed together.  

For this reason we should expect that the domain file will contain these values as static numbers for each grid cell.

Do we want to require this or should we fall back on estimating the values from the given forcings, but provide a warning to the user?

bartnijssen:
Require - this is a major problem with the current implementation of MTCLIM in VIC 4.2 and one of the reason to pull it out. It seriously messes up data assimilation and forecast projects.

bartnijssen:
If people want to reproduce the behavior of the old VIC implementation they can do so by updating the domain file with the means over the length of the model run.

arbennett:
Fixed in  4b8d37a 6

PR #66, opened 2017-03-21
Feature/annual aggregates

arbennett:
- [x] tests passed

This pull request allows for annual chunking of the output. It is a step forward in 2 directions:

1. This makes long runs less memory intensive - in tradeoff for run time.
2. We can now pull out the annual and rolling precipitation algorithms from MtClim and put them into the MetSim object. This will make restarts possible and solve #21. This is planned as an imminent pull-request.

It also brings infrastructure for chunking at different time levels, but the API/configuration for those changes needs to be addressed before it can be implemented.


arbennett:
I'm closing this until we get restarts working.

scripts/ms
```diff
@@ -35,6 +35,7 @@ def parse(args):
     parser.add_argument('config')
     parser.add_argument('-n', '--n-processes', default=1, type=int)
     parser.add_argument('-v', '--verbose', action='store_true')
+    parser.add_argument('-a', '--annual', action='store_true')
```
jhamman: did we talk about moving this to the config file where the option would be something like `grouper` and the user could specify something like `time.year` or `time.month`?

metsim/metsim.py
```diff
@@ -154,94 +158,220 @@ def load(self, job_list: list):
...
+                        t_begin = [self.state['t_min'].values[-1, i, j],
+                                   self.state['t_max'].values[-1, i, j]]
+                    try:
+                        nextday = pd.datetime(int(label)+1, 1, 1)
+                        t_end = [self.met_data['t_min'].sel(
+                                     time=nextday).isel(lat=i, lon=j),
+                                 self.met_data['t_max'].sel(
+                                     time=nextday).isel(lat=i, lon=j)]
+                    except (KeyError, ValueError):
+                        t_end = None
```
jhamman: Can you add some comments to this section. I'm a bit confused why `t_begin` and `t_end` need to be lists `[tmin, tmax]` and the try except statements are a bit tough to follow so some comments would be useful.

scripts/ms
arbennett: We talked about it when the time chunking was done through `xarray` so the options were limited by the `groupBy` function.  I think how we do it now this would be relatively easy.  I'll add an issue for this and can address it in another pull request since this one's already pretty big and complicated.

metsim/disaggregate.py
```diff
@@ -21,43 +21,51 @@
...
+        the record.
 
     Returns
     -------
     df_disagg:
         A dataframe with sub-daily timeseries.
     """
-    stop = params['stop'] + pd.Timedelta('1 days')
-    dates_disagg = pd.date_range(params['start'], stop,
+    stop = df_daily.index[-1] + pd.Timedelta('23 hours')
```
arbennett: You are correct, will fix.

metsim/methods/mtclim.py
```diff
@@ -157,42 +144,8 @@ def _calc_tfmax(prec, dtr, sm_dtr):
...
-        else:
-            window[:90] = df['prec'][:90]
-
-        parray = cnst.DAYS_PER_YEAR * window.rolling(
-            window=90, win_type=win_type, axis=0).mean()[90:]
-
-    # Convert to cm
-    parray = parray.where(parray > 80.0, 80.0) / cnst.MM_PER_CM
-    # Doing this way because parray.reindex_like(df) returns all nan
-    parray.index = df.index
```
arbennett: This is now in `metsim.py` so that we can preprocess the whole dataset using the state file.

bartnijssen:
I did not mean to approve these changes yet - not sure how to remove my approval ....

PR #79, opened 2017-06-28
Cleanup

arbennett:
- [x] tests added / passed

Fixes bugs and cleans up code for release.


metsim/io.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def read_state(params: dict, domain: xr.Dataset) -> xr.Dataset:
+    """Load in a state file"""
+    read_funcs = {
+        "netcdf": read_netcdf,
+        "data": read_data
+    }
+    start = params['start'] - pd.Timedelta('90 days')
+    stop = params['start'] - pd.Timedelta('1 days')
```
bartnijssen: Does this cover 89 days or 90?

metsim/io.py
```diff
@@ -0,0 +1,227 @@
...
+        "netcdf": read_netcdf,
+        "data": read_data
+    }
+    return read_funcs[params['forcing_fmt']](
+        params['forcing'], domain=domain, iter_dims=params['iter_dims'],
+        start=params['start'], stop=params['stop'],
+        calendar=params['calendar'], var_dict=params.get('forcing_vars', None))
+
+
+def process_vic(params: dict, domain: xr.Dataset) -> xr.Dataset:
```
bartnijssen: Is this VIC-ASCII or VIC-NetCDF? The naming is ambiguous, since the other `process_x` functions are specific to the format. If it is VIC-ASCII simply rename to `process_vic_ascii`

metsim/io.py
```diff
@@ -0,0 +1,227 @@
...
+        ds['day_of_year'] = xr.Variable(('time', ), dates.dayofyear)
+
+    if domain is not None:
+        ds.sel(**{d: domain[d] for d in iter_dims})
+    out = ds.load()
+    ds.close()
+    return out
+
+
+def read_data(data_handle, domain=None, iter_dims=['lat', 'lon'],
```
bartnijssen: `read_xarray_dataset` would be more specific

metsim/io.py
```diff
@@ -0,0 +1,227 @@
...
+    n_days = len(dates)
+    type_lookup = {'signed': 'h', 'unsigned': 'H'}
+    # Pack these for nicer syntax in the loop
+    var_names = var_dict.keys()
+    data_list = [[] for var in var_dict.keys()]
+    n_vars = len(var_names)
+    params = [p for p in var_dict.values()]
+    scales = [float(s.split()[0]) for s in params]
+    datatypes = [type_lookup[s.split()[-1]] for s in params]
+    with open(data_handle, 'rb') as f:
```
bartnijssen: I think you can speed this by reading larger chunks (VIC ascii files are sufficiently small that we should be able to read them into memory in one go), but let's leave it for now. We can change that later.

tests/test_metsim.py
```diff
@@ -176,14 +169,12 @@ def test_mtclim(test_setup):
...
     test_setup.params['out_vars'] = hourly_out_vars
-    test_setup.load()
-    assert test_setup.ready
 
     # Check to see that the data is valid
     assert type(test_setup.met_data) is xr.Dataset
 
     test_setup.run()
-    hourly = test_setup.output.isel(lat=loc[0], lon=loc[1]).to_dataframe()
+    hourly = test_setup.output.isel(lat=2, lon=2).to_dataframe()
```
bartnijssen: What's the `2` being hardwired?

metsim/io.py
arbennett: It's 90.  I just double checked to make sure.

metsim/io.py
arbennett: It's both VIC-ASCII and VIC-Binary.  In the docs I make it clear that the only VIC-specific support is from 4.2.  User's should never have to specify anything about this, though.

tests/test_metsim.py
arbennett: Had to select one of them.  The corners of this dataset are not in the mask, so it had to be greater than 0, and I was into the number 2 at the time.

PR #82, opened 2017-07-11
Fix domain subsetting, SWE melt, docstrings.

arbennett:
- [x] tests passed

### DIFF
--- a/metsim/io.py
+++ b/metsim/io.py
@@ -152,7 +152,7 @@ def read_netcdf(data_handle, domain=None, iter_dims=['lat', 'lon'],
         ds['day_of_year'] = xr.Variable(('time', ), dates.dayofyear)
 
     if domain is not None:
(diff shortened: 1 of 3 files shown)


metsim/io.py
```diff
@@ -173,7 +173,7 @@ def read_data(data_handle, domain=None, iter_dims=['lat', 'lon'],
         data_handle['day_of_year'] = xr.Variable(('time', ), dates.dayofyear)
 
     if domain is not None:
-        data_handle.sel(**{d: domain[d] for d in iter_dims})
+        data_handle = data_handle.sel(**{d: domain[d] for d in iter_dims})
```
jhamman: I think this can also be written as:

```Python
data_handle = data_handle.loc[{d: domain[d] for d in iter_dims}]
```

same in line 155. I think its just a bit cleaner not to use the `**` dict expansion.

PR #85, opened 2017-07-14
Add parameter set and variable names to output metadata

### LINKED ISSUE #24: Add parameter set to output metadata

### LINKED ISSUE #25: Add parameter for output precision

arbennett:
- [x] closes #24 
 - [x] tests added / passed

### DIFF
--- a/metsim/metsim.py
+++ b/metsim/metsim.py
@@ -39,6 +39,7 @@ from multiprocessing import Pool
 import numpy as np
 import pandas as pd
 import xarray as xr
+from collections import OrderedDict
 
 from metsim import io
 from metsim.methods import mtclim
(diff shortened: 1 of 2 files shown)


metsim/metsim.py
```diff
@@ -66,12 +67,20 @@
...
+                  'standard_name': 'air_temperature'},
+         'vapor_pressure': {'units': 'mbar', 'long_name': 'vapor pressure',
+                            'standard_name': 'vapor_pressure'},
+         'tskc': {'units': 'fraction', 'long_name': 'cloud fraction',
+                  'standard_name': 'cloud_fraction'},
+         'rel_humid': {'units': '%', 'long_name': 'relative humidity',
+                       'standard_name': 'relative_humidity'},
          '_global': {'conventions': '1.6', 'title': 'Output from MetSim',
                      'institution': 'University of Washington',
                      'source': 'metsim.py',
```
jhamman: I think it probably makes sense to make all of these OrderedDict type for consistency.

metsim/metsim.py
```diff
@@ -275,6 +281,16 @@ def setup_output(self, prototype: xr.Dataset=None):
         dims = ('time', ) + self.domain['mask'].dims
         coords = {'time': times, **self.domain.mask.coords}
         self.output = xr.Dataset(coords=coords)
+        for k, v in self.params.items():
+            # Need to convert some parameters to strings
+            if k in ['start', 'stop', 'annual', 'mtclim_swe_corr']:
+                v = str(v)
+            # Don't include complex types
+            if type(v) is list:
```
jhamman: 1) I think it is better to use `isinstance` when checking types.
2) We should check generally for `v` being an iterable. For example, a `tuple` should be treated in the same way here.

metsim/metsim.py
```diff
@@ -275,6 +281,16 @@ def setup_output(self, prototype: xr.Dataset=None):
...
+                v = ', '.join(v)
+            elif type(v) in [pd.DatetimeIndex, OrderedDict]:
+                continue
```
jhamman: so, we're just skipping these?

metsim/metsim.py
arbennett: I'm not sure I understand why this would change anything in the output?

metsim/metsim.py
arbennett: I take back my thumbs up.  Strings are iterable...

jhamman:
```Python
if not isinstance(v, str) and hasattr(v, '__iter__'):
    # or similar
```

?

metsim/metsim.py
jhamman: Just so we get consistent ordering in the output file.

arbennett: Fixed in lines below.

PR #88, opened 2017-08-07
Feature/state filename

### LINKED ISSUE #87: Add config file option for defining the statefile name
The current behavior of MetSim's state file write is to:

```Python
self.state.to_netcdf(os.path.join(self.params['out_dir'], 'state.nc'),
                     encoding={'time': {'dtype': 'f8'}})
```

I'm proposing a new config file option that allows users to specify the output filename/filepath in addition to the input filename for the statefile.

jhamman:
- [x] closes #87
 - [x] tests passed
 - [x] passes ``git diff upstream/develop | flake8 --diff``


bartnijssen:
Tests don't pass (travis tests) - I think this is just a problem with path names in the tests themselves. I unchecked the `tests passed` box

PR #93, opened 2017-08-11
added option to use forcing start/stop dates and misc improvements to…

### LINKED ISSUE #92: Additional time handling functionality
The most common error I've run into when running MetSim is related to ill-specified time values of one kind or another. This could be in the state file, configuration file, or in the forcing file. All this leads me to think we could use 

1) some more robust validation of the time variables in MetSim,
2) some more instructive errors when things go wrong, and
3) a new feature to infer run length from the forcing file.

The underlying goal is to make the running/output of MetSim predictable with regards to the specification of input/output times.

cc @arbennett: following up on our personal conversation yesterday

jhamman:
… time checking / loading of data

 - [x] closes #92
 - [x] tests added / passed
 - [ ] whatsnew entry

### DIFF
--- a/metsim/datetime.py
+++ b/metsim/datetime.py
@@ -58,11 +58,12 @@ def date_range(start=None, end=None, periods=None, freq='D', tz=None,
             start_num, end_num = date2num(
                 pd.to_datetime([start, end]).to_pydatetime(),
                 units, calendar=calendar)
-            periods = int((end_num - start_num) / steps)  # Todo divide by freq
+            # Todo divide by freq
+            periods = int((end_num - start_num) / steps) + 1
 
             times = num2date(
                 np.linspace(start_num, end_num, periods,
-                            endpoint=False,
+                            endpoint=True,
                             dtype=np.float128), units, calendar)
 
             index = pd.DatetimeIndex(xr.conventions.nctime_to_nptime(times))
@@ -89,5 +90,7 @@ def units_from_freq(freq, origin=DEFAULT_ORIGIN):
         return 'hours since %s' % origin
     elif 'D' in freq:
         return 'days since %s' % origin
+    elif 'T' in freq:
+        return 'minutes since %s' % origin
     else:
-        raise NotImplementedError()
+        raise NotImplementedError('freq %s not supported at this time' % freq)
--- a/metsim/datetime.py
+++ b/metsim/datetime.py
@@ -93,4 +93,5 @@ def units_from_freq(freq, origin=DEFAULT_ORIGIN):
     elif 'T' in freq:
         return 'minutes since %s' % origin
     else:
-        raise NotImplementedError('freq %s not supported at this time' % freq)
+        raise NotImplementedError(
+            'freq {} not supported at this time'.format(freq))
--- a/metsim/io.py
+++ b/metsim/io.py
@@ -63,11 +63,11 @@ def read_state(params: dict, domain: xr.Dataset) -> xr.Dataset:
         "netcdf": read_netcdf,
         "data": read_data
     }
-    start = params['start'] - pd.Timedelta('90 days')
-    stop = params['start'] - pd.Timedelta('1 days')
+
     return read_funcs[params['state_fmt']](
         params['state'], domain=domain, iter_dims=params['iter_dims'],
-        start=start, stop=stop, calendar=params['calendar'],
+        start=params['state_start'], stop=params['state_stop'],
+        calendar=params['calendar'],
         var_dict=params.get('state_vars', None))
 
 
@@ -146,18 +146,18 @@ def read_netcdf(data_handle, domain=None, iter_dims=['lat', 'lon'],
     ds = xr.open_dataset(data_handle)
 
     if var_dict is not None:
+        var_list = list(var_dict.keys())
+        ds = ds[var_list]
         ds.rename(var_dict, inplace=True)
 
-    if start is not None and stop is not None:
+    if start is not None or stop is not None:
         ds = ds.sel(time=slice(start, stop))
         dates = ds.indexes['time']
         ds['day_of_year'] = xr.Variable(('time', ), dates.dayofyear)
 
     if domain is not None:
         ds = ds.sel(**{d: domain[d] for d in iter_dims})
-    out = ds.load()
-    ds.close()
-    return out
+    return ds
 
 
 def read_data(data_handle, domain=None, iter_dims=['lat', 'lon'],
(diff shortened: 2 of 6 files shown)


metsim/datetime.py
```diff
@@ -89,5 +90,7 @@ def units_from_freq(freq, origin=DEFAULT_ORIGIN):
         return 'hours since %s' % origin
     elif 'D' in freq:
         return 'days since %s' % origin
+    elif 'T' in freq:
+        return 'minutes since %s' % origin
     else:
-        raise NotImplementedError()
+        raise NotImplementedError('freq %s not supported at this time' % freq)
```
arbennett: I'd prefer we try to use `.format(...)` as much as possible.  It looks like there's a couple uses of the `%` formatting above as well that I didn't catch before.

metsim/io.py
```diff
@@ -146,18 +146,18 @@ def read_netcdf(data_handle, domain=None, iter_dims=['lat', 'lon'],
     ds = xr.open_dataset(data_handle)
 
     if var_dict is not None:
+        var_list = list(var_dict.keys())
+        ds = ds[var_list]
         ds.rename(var_dict, inplace=True)
 
-    if start is not None and stop is not None:
+    if start is not None or stop is not None:
```
arbennett: Doesn't this have the chance to break? For example, if `start=None` and `stop!=None` we will end up with (in the line below) `ds = ds.sel(time=slice(None, stop))`.

metsim/metsim.py
```diff
@@ -148,12 +148,84 @@ def __init__(self, params: dict):
...
         self.met_data = io.read_met_data(self.params, self.domain)
+        self._validate_force_times(force_times=self.met_data['time'])
+        logger.info("read_state")
         self.state = io.read_state(self.params, self.domain)
         self.met_data['elev'] = self.domain['elev']
         self.met_data['lat'] = self.domain['lat']
+        logger.info("_aggregate_state")
         self._aggregate_state()
+        logger.info("load_inputs")
+        self.load_inputs()
```
arbennett: I'm wary of having `self.load_inputs()` after `self._aggregate_state()`.  The whole reason we were doing the `load` and `close` in the IO routines was to avoid the situation where the state and input forcings are looking at the same file and undefined behavior occurred.  It might be possible to reason through and prove that this situation won't occur with this setup, but I think it's easier to side with caution on this one.

metsim/metsim.py
```diff
@@ -148,12 +148,84 @@ def __init__(self, params: dict):
...
+                    # end from config file
+                    year, month, day = self.params[p].split('/')
+                    self.params[p] = pd.datetime(int(year), int(month),
+                                                 int(day))
+                else:
+                    self.params[p] = pd.to_datetime(self.params[p])
+            else:
+                self.params[p] = pd.to_datetime(self.params[p])
+
+        logger.info('start %s' % self.params['start'])
```
arbennett: String formatting again. (And below)

metsim/metsim.py
```diff
@@ -307,8 +379,11 @@ def setup_output(self, prototype: xr.Dataset=None):
             # Need to convert some parameters to strings
             if k in ['start', 'stop', 'annual', 'mtclim_swe_corr']:
                 v = str(v)
+            elif k in ['state_start', 'state_stop']:
+                # skip
+                continue
             # Don't include complex types
-            if isinstance(v, dict):
+            elif isinstance(v, dict):
```
arbennett: I think I like this being an `if` more, since it's talking about a separate variable than the first `if` statement.  This allows for us to catch more than one situation, and is just a more cautious approach in general.

metsim/metsim.py
```diff
@@ -326,31 +401,37 @@ def setup_output(self, prototype: xr.Dataset=None):
     def _aggregate_state(self):
         """Aggregate data out of the state file and load it into `met_data`"""
         # Precipitation record
-        begin_record = self.params['start'] - pd.Timedelta("90 days")
-        end_record = self.params['start'] - pd.Timedelta("1 days")
-        record_dates = date_range(begin_record, end_record,
+
+        assert self.state.dims['time'] == 90
```
arbennett: I really don't like this check, because it seems to limit some cases that are pretty useful in my mind.  For example, if you have a long timeseries broken up into 1 year chunks you could previously just use the previous year's data as your state file.  Now it looks like you'd have to preprocess all of those files and create an additional state file for each years input.

metsim/io.py
jhamman: `ds = ds.sel(time=slice(None, stop))` will work. Just like this works:

```python
In [1]: x = list(range(10))

In [2]: x
Out[2]: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]

In [3]: x[slice(None, 4)]
Out[3]: [0, 1, 2, 3]
```

metsim/metsim.py
jhamman: I'm skeptical that this is actually a problem. This is all happening before we startup MetSim's multiprocessing which is where I'd image any odd behavior came from. The reason I like this method is that we fully define exactly what data we need in the previous lines before we ever load any data. I think this is preferable to blinding loading data we don't know if we need.

metsim/metsim.py
jhamman: Not really, our `read_state` function handles this correctly now with improved slice selection.

metsim/io.py
arbennett: Interesting, I had no idea that you could do that.

metsim/datetime.py
jhamman: Done, though the logging module doesn't require the `.format` method to be called. This works:

```python
logger.info('just a {} example', 'little')
```

xref: https://docs.python.org/3/howto/logging-cookbook.html#using-custom-message-objects

metsim/metsim.py
jhamman: Done. I misread this logic. Good catch.

metsim/metsim.py
jhamman: I'm open to other ideas for how to pin down the behavior in this method but I was finding it useful to explicitly check the state shape before going into the computations in this method.

arbennett: Ah, I didn't catch that's what you were doing with the `state_start` and all that. In that case, I like this.

metsim/metsim.py
arbennett: That's a good point, and I think it's reasonable to say that no issues will occur after looking at it a bit more. I'm still a bit sensitive to that bug because it took so long to track down.

PR #102, opened 2017-09-28
Improve setup validation

arbennett:
- [x] tests added / passed

These changes make it more clear what variables must be specified in the input parameter set by checking the setup before running, as well as using default parameters to catch errors.


metsim/metsim.py
```diff
@@ -483,9 +486,9 @@ def _validate_setup(self):
         out_var_check = ['temp', 'prec', 'shortwave', 'vapor_pressure',
                          'air_pressure', 'rel_humid', 'spec_humid',
                          'longwave', 'tsck', 'wind']
-        if int(self.params['time_step']) == 1440:
+        if int(self.params.get('time_step', -1)) == 1440:
```
jhamman: what do you think about using `None` as the sentinel value instead?

arbennett: Only reason I went with `-1` here was because I did it up at line 468 so I could modulus it.  I'm fine with either, but think there's a case for consistency.

jhamman: I think either is fine but if we're going to use `-1` in multiple places, how about a global `MISSING` variable?

arbennett: I'm not a big fan of that idea because it ends up being an abstraction that only works in some contexts.

PR #105, opened 2017-10-07
Feature/wind output

### LINKED ISSUE #99: Add wind to the sample datasets
This is so that the end result of MetSim will be consistent with an input into VIC.  Otherwise, a new VIC user might assume that one should just run MetSim as shown in the sample dataset, and then get an error when attempting to run VIC with those outputs.

For pressure, the fix is very simple; just specify pressure as an output variable in the sample config file.
For wind, someone will have to put reasonable daily wind values into the input daily meteorology file.   Here at ASU we might be able to do that.

The VIC error when wind is not present is also not very informative (it says there's an error reading a variable from the forcing file, but because the config file was set up to cover only those variables that were actually in the forcing file, which lacked wind, the VIC error message didn't actually know the name of the wind variable).  I realize this is a VIC issue rather than a MetSim issue.  But still, this is the type of situation that will arise.

### LINKED ISSUE #108: README.md minor inaccuracies
The version of `README.md` in the develop branch has a few minor inaccuracies:
1. In section 2, it lists the set of default daily variables.  This list doesn't match how the `out_vars` list in the `params` dictionary is initialized.  Shouldn't it?
2. Currently, `swe` is listed as one of those default variables, but is no longer in the `out_vars` list since PR #103.  So perhaps it is worth also listing those variables that are not output by default but are optionally available? 
3. Given that PR #105 adds `wind` and `pressure` to the `out_vars` list, PR #105 should probably add those variables to `README.md`.  They will become defaults, so regardless of the answer to point 2, they belong in `README.md`.

kwhitney727:
- [x] closes #99
 - [x] closes #108
 - [x] tests added / passed
 - [ ] whatsnew entry

Not sure how to resolve remaining messages from ``git diff upstream/master | flake8 --diff``

The changes made here adds "wind" to the forcing output, since this is required for the VIC model.


jhamman:
@kwhitney727 / @tbohn - can you rebase this to only include the wind changes? This can go into the 1.2 release.

tbohn:
But PR #104 is ready to go into 1.2 as well.  @arbennett suggested changes in his review and we made them.  Haven't heard back since - perhaps people have forgotten about it?  PR #104 is the most important of our PRs.

arbennett:
This is superceded by #155 - closing.

PR #106, opened 2017-10-07
Feature/time wrap

kwhitney727:
- [ ] closes #xxxx
 - [x] tests added / passed
 - [ ] whatsnew entry

Not sure how to resolve remaining messages from ``git diff upstream/master | flake8 --diff``

A new script, "wrap_metsim_over_time.py", has been added. This is a script that wraps through a directory of monthly forcing files, running metsim for each file using the state file output of the previous month's run. The script concatenates all monthly outputs to either a single yearly output file for the case where there are a year or more worth of input forcings, or a single file with all months' outputs for the case where there is less than a years worth of input forcings.


jhamman:
I think we can close this based on #127. Any objections?

tbohn:
I haven't had time to test out the new code from #127, but assuming it accomplishes the same thing that this PR does (multiple input and output files), I have no objection.

arbennett:
This type of functionality has been implemented in what will be 2.0.

PR #113, opened 2018-01-24
[WIP] Feature/dask

### LINKED ISSUE #101: Refactor metsim parallelization
Xarray 0.10 will include parallelized array rolling means and a `to_dask_dataframe` methods. This should allow us to fairly easily refactor the metsim parallelization to better leverage these tools with the goal of getting better parallel performance with a smaller memory footprint.

jhamman:
- [x] closes #101
 - [ ] tests added / passed
 - [ ] whatsnew entry


metsim/methods/mtclim.py
```diff
@@ -19,197 +19,125 @@
...
-
-    # Calculate cloud effect
+    tdew_old = t_min
+
+    while(np.sqrt(np.mean((tdew_temp-tdew_old)**2)) > params['tdew_tol']):
+        tdew_old = tdew_temp.copy()
+        vp = svp(tdew_temp)
+        sw = calc_shortwave(yday, tfmax, vp, sg)
+        pet = _calc_pet(sw, t_day, dayl, params)
+        tdew_temp = calc_tdew(pet, t_min, seasonal_prec, dtr)
```
jhamman: this function needs more work, previously the RMSE was calculated over the full timeseries so we'll need to work that in here.

metsim/methods/mtclim.py
```diff
@@ -19,197 +19,125 @@
...
+def _calc_pet(shortwave, t_day, dayl, params):
+    pa = atm_pres(params['elev'], params['lapse_rate'])
+    return calc_pet(shortwave, t_day, dayl, pa) * cnst.MM_PER_CM
+
+
+def calc_tdew(pet, t_min, seasonal_prec, dtr):
+    ratio = pet / seasonal_prec
+    return ((t_min + cnst.KELVIN) * (-0.127 + 1.121 * (1.003 - 1.444 * ratio
+            + 12.312 * np.power(ratio, 2) - 32.766 * np.power(ratio, 3))
+            + 0.0006 * dtr) - cnst.KELVIN)
```
jhamman: we may need to create a xarray ufunc for `np.power`

metsim/metsim.py
```diff
@@ -257,80 +253,15 @@ def launch(self):
...
-                    self.state.sel(**locs), self.disagg, times, label)
-
-                # Cut the returned data down to the correct time index
-                self._unpack_results(wrap_results)
-
-            self.write(label)
+        if self.disagg:
+            self.output = disaggregate(out)
+        else:
+            self.output = out
```
jhamman: More logic to come here. Specifically, we don't have any notion of chunked output (per year files). We also don't have any logic yet for specifying dask chunking schemes.

scripts/ms
```diff
@@ -53,6 +58,19 @@ def parse(args):
...
+        from dask.distributed import Client
+        client = Client(scheduler_file=opts.scheduler_file)
+        print(client)
+    elif opts.nthreads > 1 or opts.n_procs > 0:
+        from dask.distributed import Client
+        client = Client(n_workers=opts.nprocs, n_threads=opts.nthreads)
+        print(client)
+    else:
+        client = None
+
```
jhamman: This is maybe a bit verbose but allows for a range of dask schedular configurations. The `schedular_file` option would allow for the user to use a previously launched dask cluster with many options. The nthreads/nprocs concept would launch a local cluster.

metsim/metsim.py
tbohn: @jhamman just FYI, Kristen Whitney here at ASU has created a python script to loop over years and call MetSim on individual monthly forcing files and concatenate the output into yearly files...  It's PR #106 ...  Will the dask feature supercede this?

jhamman:
This PR is being superseded by #125.

metsim/metsim.py
jhamman: @tbohn - I think we can handle that in #125.

Issue #114, opened 2018-01-25
Handle case where number of sunrises != number of sunsets


arbennett:
When UTC offsets are added (solving #9), it becomes pretty much a certainty that there will be grid cells that do not end up with the same number of sunrises as sunsets, meaning that `set_min_max_times` will fail.  This can probably be accounted for by taking the mean time between sunrise and sunset of the ones that do occur and use that to estimate when the next one _would've_ occurred. Then, we will have to interpolate over a longer timeperiod and slice off just the time period we are interested in.

arbennett:
Solved in #120

PR #120, opened 2018-02-15
Refactor/disagg

### LINKED ISSUE #9: Provide better handling of time zones
After discussion of #8 for adjusting time within a given timezone the issue came up that we may want to account for different time zone standards.  The current implementation is similar to nautical time, with 24 equally spaced time zones (15 degree longitude widths).

A situation where this might not be the best is if forcing data is provided from a source where timezones are given in some other format, where there would be a discrepancy between the the expected offset and the actual offset.

arbennett:
- [x] closes #9, #114

This is a cleanup of the disaggregation routines, which includes an implementation for UTC offsets in the output.  This is a breaking change, since we can no longer match VIC to the previously required degree of accuracy (mainly on temperature disaggregation). Like the `mtclim` refactor, this moves all functionality from Pandas datastructures to Numpy ones, in order to aid the revamp of parallelism.

This also enforces `t_min < t_max` on all input data.


metsim/methods/mtclim.py
```diff
@@ -19,197 +19,80 @@
...
-                    1.0e6 / daylength[yday][inds])
-        sc = np.maximum(sc, 100.)
-
-    # Calculation of shortwave is split into 2 components:
-    # 1. Radiation from incident light
-    # 2. Influence of snowpack - optionally set by MTCLIM_SWE_CORR
-    df['shortwave'] = potrad[yday] * t_final + sc
-
-    # Calculate cloud effect
+def run(df, params):
```
jhamman: @arbennett - do you know why the mtclim changes are here in this PR?

jhamman: Actually, it looks like there are some conflicts to resolve. Let me know once those are fixed.

PR #124, opened 2018-02-23
Fix/time metadata

### LINKED ISSUE #123: Time metadata not updated for UTC offsets

arbennett:
- [x] closes #123
 - [x] tests added / passed


metsim/metsim.py
```diff
@@ -225,6 +224,9 @@ def _validate_force_times(self, force_times):
         logger.info('state stop {}'.format(self.params['state_stop']))
 
         logger.info('calendar {}'.format(self.params['calendar']))
+        if self.params['utc_offset']:
+            attrs['time'] = {'long_name': 'UTC time',
+                  'standard_name': 'utc_time'}
```
jhamman: rather than special casing  when `utc_offset = True`, can we remove the `time` attributes from the `attrs` dict and set both cases here. I think that is going to be easier to understand for future readers.

PR #125, opened 2018-02-24
Feature/dask2

### LINKED ISSUE #101: Refactor metsim parallelization

jhamman:
- [x] closes #101, #113
 - [ ] tests added / passed
 - [x] passes ``git diff upstream/develop | flake8 --diff``
 - [ ] whatsnew entry

This approach parallelizes metsim using a series of delayed functions. It also implements some new command line interfaces to make it easy to switch between dask schedulers.

cc @arbennett for the first review.


scripts/ms
```diff
@@ -82,28 +63,79 @@ def init(opts):
...
     conf['iter_dims'] = to_list(conf.get('iter_dims', '["lat", "lon"]'))
-    if opts.time_grouper is not None:
-        conf['time_grouper'] = opts.time_grouper
+    conf['time_grouper'] = time_grouper if time_grouper else None
     conf = {k: v for k, v in conf.items() if v != []}
     return conf
 
 
-def main():
+def setup_dask(scheduler, num_workers):
```
jhamman: @arbennett - I'd like to test this out for each scheduler with varying number of workers using the benchmarking dataset you had used previously. Can you point me to that dataset?

scripts/ms
```diff
@@ -82,28 +63,79 @@ def init(opts):
...
+    return dask.set_options(**options)
+
+
+@click.command(context_settings=dict(help_option_names=['-h', '--help']))
+@click.argument('config', type=click.Path(exists=True))
+@click.option('--scheduler', type=str, default=None,
+              help='Parallel dask scheduler to use, defaults to '
+                   'multiprocessing when num_workers is set.')
+@click.option('-n', '--num_workers', default=None, type=int,
+              help='Number of workers to use during parallel computations')
```
jhamman: Need to look into this. It is possible that the default is being typecast to an int which is breaking some of the default behavior in `setup_dask`.

metsim/metsim.py
```diff
@@ -263,34 +263,41 @@ def launch(self):
...
-            self.pool.close()
-            # Check that everything worked
-            [stat.get() for stat in status]
-            self.pool.join()
+                delayed_objs.append(run_func(self.method.run, locs,
+                                             self.params,
+                                             data.sel(**locs),
+                                             self.state.sel(**locs),
+                                             self.disagg,
+                                             times, label).persist())
```
jhamman: I'm considering unpacking the inputs to this function from their xarray data structures into something simpler (dict of numpy arrays). I'm hypothesizing that this will make it easier to serialize the data when using multiprocessing. @arbennett, thoughts?

jhamman: My simple tests indicate that this may yield about a 2x speedup.

scripts/ms
```diff
@@ -82,28 +63,79 @@ def init(opts):
...
+@click.option('-tg', '--time_grouper', default=None, type=str, multiple=True,
+              help='Pandas TimeGrouper string (e.g. `1AS`)')
+@click.option('-v', '--verbose', count=True,
+              help='Set the Verbosity of MetSim')
+def main(config, scheduler, num_workers, time_grouper, verbose):
```
jhamman: @arbennett - any comments on this user interface?

metsim/metsim.py
arbennett: This seems like a good idea to me.

scripts/ms
arbennett: I think it might be useful to list out the main schedulers available for use in the help, but other than that I like it.

metsim/metsim.py
```diff
@@ -263,34 +263,41 @@ def launch(self):
         time_dim, groups = self._get_time_dim_and_time_dim()
 
         for label, times in groups.items():
-            self.pool = Pool(processes=nprocs)
-            self.site_generator = itertools.product(*iter_list)
+            self.site_generator = list(itertools.product(*iter_list))
```
arbennett: Is there a reason we need to compute this right away? Is it to use the `progress` function?

arbennett: Also, it might be worth renaming to `site_list` since it's no longer a generator.

metsim/metsim.py
```diff
@@ -263,34 +263,41 @@ def launch(self):
...
                 # Don't run masked cells
                 if not self.domain['mask'].sel(**locs).values > 0:
                     continue
-                stat = self.pool.apply_async(
-                    wrap_run,
-                    args=(self.method.run, locs, self.params,
-                          data.sel(**locs),
-                          self.state.sel(**locs),
-                          self.disagg, times, label),
-                    callback=self._unpack_results)
```
arbennett: We can probably get rid of the `_unpack_results` method now, and just do the explicit unpacking in the single core version.

scripts/ms
jhamman: done in 08f8199

PR #127, opened 2018-03-08
Feature/open mfdataset

### LINKED ISSUE #126: ENH: use open_mfdataset when reading netcdf files
It would be nice if the input forcings to metsim (`forcing` config option) was allowed to be a glob-like file path. This, combined with xarray.open_mfdataset() will allow for the input of multiple forcing files, munged together via xarray.

jhamman:
- [x] closes #126
 - [x] tests passed
 - [x] whatsnew entry

### DIFF
--- a/docs/whats-new.rst
+++ b/docs/whats-new.rst
@@ -18,7 +18,9 @@ Enhancements
 - Improved configuration validation by checking for correctness of output variables (:issue:`96`)
   By `Andrew Bennett <https://github.com/arbennett>`
 - Added option to skip reading ``swe`` variable from state file if it is not
-  going to be used by MtClim. (:issue:`XX`). By `Joe Hamman <https://github.com/jhamman>`_.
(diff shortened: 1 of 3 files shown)


arbennett:
Can you give an example of how this would work in a configuration file, or is this mainly for scripting purposes?

jhamman:
What I'd like to be able to do in the configuration file is either:

```yaml
forcing = "/path/to/forcing/files/forcings_by_year*nc"
```

Where the wildcard is expanded to a selection of files. Or provide a list of inputs files that xarray can concatenate/merge:

```yaml
forcing =  ["/path/to/forcing/files/forcings_by_year_2000.nc", "/path/to/forcing/files/forcings_by_year_2001.nc"]
```

arbennett:
Thanks, I think supporting both is a good idea.

jhamman:
The changes in this PR enable both, while at the same time still supporting a single file as an input.

PR #128, opened 2018-03-27
default to f4 and don't override encoding

jhamman:
@arbennett - I was unable to get the `out_precision` option to work prior to these changes. I also want to advocate for a change in the default precision to `f4`, mostly because its mtclim 😉.


jhamman:
I'm less passionate about the suggested change to the default precision now than I was 2 months ago but the bug fix in here should still move forward. Let me know if you want me to revert the change to the precision.

arbennett:
Personally, I'm fine with either output precision - it's possible to set in the configuration either way.

PR #135, opened 2018-05-29
Feature/versioneer

jhamman:
- [ ] closes #xxxx
 - [ ] tests added / passed
 - [ ] whatsnew entry


arbennett:
I have a general question about this PR.  Why is the `versioneer.py` module here?  Is this automatically generated when you run `versioneer install` per the readme (https://github.com/warner/python-versioneer)?

jhamman:
Yes, exactly. This PR will allow us to skip the step of manually updating the version number in setup.py for each release. Moreover, it allows metsim's version string to reflect development state and/or a dirty state. Let me add another document that describes the release steps.

jhamman:
@arbennett - see updated PR. The idea here is make making a release as easy as creating a tag. Versioneer is just the easiest way I know of to manage updating the version info.

PR #138, opened 2018-05-30
include license in manifest.in

jhamman:
- [ ] closes #xxxx
 - [x] tests added / passed
 - [ ] whatsnew entry


jhamman:
I think this is technically required by the GPL license and required a work around in the conda-forge recipe.

PR #147, opened 2018-09-06
New parallel

jhamman:
- [ ] closes #xxxx
 - [ ] tests added / passed
 - [ ] whatsnew entry


metsim/metsim.py
```diff
@@ -133,34 +141,43 @@ class MetSim(object):
...
-        self._validate_setup()
-        logger.info("read_domain")
-        self.domain = io.read_domain(self.params)
+        # set global dask scheduler
+        if domain_slice is NO_SLICE:
+            if 'distributed' == self.params['scheduler']:
+                from distributed import Client
+                self._client = Client(
+                    n_workers=self.params['num_workers'], threads_per_worker=1, diagnostics_port=5555)
+                print(self._client)
```
jhamman: if scheduler is a file, we should pass scheduler to `Client(sheduler_file=scheduler)`.

metsim/metsim.py
```diff
@@ -209,169 +226,193 @@ def _validate_force_times(self, force_times):
...
+    @property
+    def met_data(self):
+        if self._met_data is None:
+            print("read_met_data")
+            self._met_data = io.read_met_data(self.params, self.domain).isel(
+                **self._domain_slice)
+            self._met_data['elev'] = self.domain['elev']
+            self._met_data['lat'] = self.domain['lat']
+            self._validate_force_times(force_times=self._met_data['time'])
+        return self._met_data
```
jhamman: we need to clip to the shape of the domain here (like we currently do in read_netcdf)

metsim/metsim.py
```diff
@@ -209,169 +226,193 @@ def _validate_force_times(self, force_times):
...
+        # TODO: we should be able to pickle the Metsim and not send the params
+        # dict in this way. This feels a bit hokey.
+        print('running chunks')
+        times = self._get_output_times()
+        filename = self._get_output_filename(times)
+        self.setup_netcdf_output(filename, times)
+
+        # write_lock = _get_scheduler_lock(self.params['scheduler'], path_or_file=filename)
+        m = multiprocessing.Manager()
+        write_lock = m.Lock()
```
jhamman: we should go back to trying the `_get_scheduler_lock` approach - I'm still not clear if we need the multiprocessing lock directly.

metsim/metsim.py
```diff
@@ -209,169 +226,193 @@ def _validate_force_times(self, force_times):
...
+            self._domain.close()
+            self._met_data.close()
+            self._state.close()
+
+    def setup_netcdf_output(self, filename, times):
+
+        from netCDF4 import Dataset
+        from cftime import date2num
+        print('debug', 'creating %s' % filename)
+        ncout = Dataset(filename, mode="w")
```
jhamman: use context manager for this file handle

metsim/metsim.py
```diff
@@ -209,169 +226,193 @@ def _validate_force_times(self, force_times):
...
-        else:
-            groups['total'] = index
+        # set variable attrs
+        for varname in ncout.variables:
+            for key, val in attrs.get(varname, {}).items():
+                setattr(ncout.variables[varname], key, val)
 
-        return index, groups
+        ncout.sync()
+        ncout.close()
```
jhamman: shouldn't need these if we add a context manager

metsim/metsim.py
```diff
@@ -209,169 +226,193 @@ def _validate_force_times(self, force_times):
...
-            self.pool.close()
-            # Check that everything worked
-            [stat.get() for stat in status]
-            self.pool.join()
-            self.write(label)
+        times = self._get_output_times()
+        filename = self._get_output_filename(times)
 
-    def run(self):
+        ncout = Dataset(filename, mode="a")
```
jhamman: context manager here, plus open with `r+`

metsim/metsim.py
```diff
@@ -209,169 +226,193 @@ def _validate_force_times(self, force_times):
...
+        times = self.met_data['time']
+
+        params = self.params.copy()
+        for index, mask_val in np.ndenumerate(self.domain['mask'].values):
+            locs = {d: i for d, i in zip(self.domain['mask'].dims, index)}
+            print(locs)
+            # Don't run masked cells
+            if not mask_val > 0:
+                print('continuing', flush=True)
+                continue
```
jhamman: turn this into an if/else. we don't need to do anything unless mask_val is positive (i.e. don't create locs)

metsim/metsim.py
```diff
@@ -511,87 +567,85 @@ def _validate_setup(self):
...
+    # def write(self, suffix=""):
+    #     """
+    #     Dispatch to the right function based on the configuration given
+    #     """
+    #     dispatch = {
+    #         'netcdf': self.write_netcdf,
+    #         'ascii': self.write_ascii,
+    #         'data': self.write_data
+    #     }
+    #     dispatch[self.params.get('out_fmt', 'netcdf').lower()](suffix)
```
jhamman: thoughts on what do with the existing writers? If it was up to me, we'd only support netcdf output.

metsim/metsim.py
```diff
@@ -690,4 +739,40 @@ def wrap_run(func: callable, loc: dict, params: dict,
...
+
+@dask.delayed()
+def wrap_run_slice(params, write_lock, domain_slice=NO_SLICE):
+    print('running chunk %s' % domain_slice, flush=True)
+    os.environ["HDF5_USE_FILE_LOCKING"] = HDF5_USE_FILE_LOCKING
+    ms = MetSim(params, domain_slice=domain_slice)
+    # TODO: determine if this lock/block is actually necessary.
+    # lock = SerializableLock()
+    # with lock:
+    #     print("load_inputs")
```
jhamman: After thinking about this a bit more, I think we should still protect with a threading lock here.

metsim/metsim.py
```diff
@@ -690,4 +739,40 @@ def wrap_run(func: callable, loc: dict, params: dict,
...
+    def right(chunk, dim):
+        nums = np.arange(chunk, dim + chunk, chunk)
+        nums[-1] = dim + 1
+        return nums
+
+    slices = [[slice(*a) for a in (zip(left(int(chunks[dim]), dims[dim]),
+                                       right(int(chunks[dim]), dims[dim])))]
+              for dim in chunks.keys()]
+
+    return [dict(zip(chunks.keys(), p)) for p in itertools.product(*slices)]
```
jhamman: write a few basic unit tests for this function.

PR #149, opened 2018-09-14
Add Zenodo author list

arbennett:


jhamman:
I did not know you could do this but I like it.

PR #156, opened 2019-02-25
Updated citation to Bohn et al (2019) and removed PtriangleMethod.pdf.

tbohn:
Updated citation to Bohn et al (2019) and removed PtriangleMethod.pdf.


tbohn:
Oops, wrong branch!

PR #163, opened 2019-04-02
fix for missing last year in output files

jhamman:
- [ ] closes #xxxx
 - [ ] tests added / passed
 - [ ] whatsnew entry

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -3,7 +3,7 @@ channels:
   - conda-forge
 dependencies:
   - python>=3.5
-  - xarray=0.10.9
+  - xarray>=0.11.0
   - pandas
   - netcdf4>=1.2.5
   - numba
@@ -12,3 +12,4 @@ dependencies:
   - dask
   - distributed
   - toolz
+  - pytest
--- a/metsim/io.py
+++ b/metsim/io.py
@@ -120,7 +120,7 @@ def read_ascii(data_handle, domain=None,
     dates = date_range(start, stop, calendar=calendar)
     names = var_dict.keys()
     ds = pd.read_csv(data_handle, header=None, delim_whitespace=True,
-                     sep='\t', names=names).head(len(dates))
+                     names=names).head(len(dates))
     ds.index = dates
     return ds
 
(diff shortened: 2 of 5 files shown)


jhamman:
@arbennett - 

I was having a problem where the last output file of my metsim runs was not written (config below). This seems to have fixed things but we should think about what we were trying to do before:


```
# This is an example of an input file for MetSim
[MetSim]
out_vars = ['temp', 'prec', 'shortwave', 'longwave', 'vapor_pressure', 'wind', 'air_pressure']

# Time step in minutes
time_step = 60
utc_offset = False  # Will need to change for SUMMA

# Forcings begin here (year/month/day:hour) (hour optional)
start = 1955/1/1

# Forcings end at this date (year/month/day)
stop = 2005/12/31

# Input and output directories
forcing = /gpfs/flash/jhamman/storylines_test/GFDL-CM3/hist/loca/downscaling_data/downscale.GFDL-CM3.hist.loca.nc
domain  = /gpfs/fs1/work/jhamman/storylines/storylines_test_data/model_inputs/domain.vic.conus0.0125deg_newman.20180326.nc
state =   /gpfs/flash/jhamman/storylines_test/GFDL-CM3/hist/loca/downscaling_data/downscale.GFDL-CM3.hist.loca.nc
forcing_fmt = netcdf
domain_fmt = netcdf
state_fmt = netcdf
out_dir = /gpfs/flash/jhamman/storylines_test/GFDL-CM3/hist/loca/disagg_data
output_prefix = forcing.GFDL-CM3.hist.loca.metsim.60
out_precision = f4
out_freq = AS
in_format = netcdf
out_fmt = netcdf
out_state = /gpfs/flash/jhamman/storylines_test/GFDL-CM3/hist/loca/disagg_data/state.GFDL-CM3.hist.loca.metsim.nc
```

arbennett:
@jhamman - I vaguely remember adding the `[:-1]` part here because I think we were choosing the date grouping in a different fashion. This was before a lot of cleanup, so I think it was probably from a logic error in something like a pandas timedelta. Since this passes tests and seems to produce what you think is correct I am happy to include it.

arbennett:
This was implemented at some point in develop, and is no longer necessary

Issue #169, opened 2019-08-29
Fail early and cleanly if IO paths do not exist


bartnijssen:
don't launch into the computations if IO paths do not exist or are not accessible. Provide a meaningful error rather than a full back trace and lots of broken pipes

PR #170, opened 2019-08-30
Allow for constant fields in MetSim input

### LINKED ISSUE #160: add option to add a constant wind field to output when not in input
For some applications wind is a secondary met input, and is required but can be specified as a constant, e.g., 1.5 m/s.   At present this must be tacked onto output after the fact, but can easily be generated internally and added to the output variables while MetSim is being run.  Suggest adding a constant wind option that allows user to specify a wind speed to be added to all files for all timesteps.  Alternatively the wind speed could be supplied as a domain variable, allowing for different wind speeds in different locations.

bartnijssen:
- [x] closes #160
 - [x] tests added / passed
 - [x] whatsnew entry

#
Merge PR #172 first

### DIFF
--- a/metsim/cli/ms.py
+++ b/metsim/cli/ms.py
@@ -63,13 +63,13 @@ def init(opts):
     def to_list(s):
         return json.loads(s.replace("'", '"').split('#')[0])
 
-
-    conf['constant_vars'] = OrderedDict(config['constant_vars'])
     conf['forcing_vars'] = OrderedDict(config['forcing_vars'])
-    # remove constant_vars from forcings_vars
-    del_keys = list(set(conf['forcing_vars'].keys()).
-                    intersection(conf['constant_vars'].keys()))
-    [conf['forcing_vars'].pop(x) for x in del_keys]
+    if 'constant_vars' in config:
+        conf['constant_vars'] = OrderedDict(config['constant_vars'])
+        # remove constant_vars from forcings_vars
+        del_keys = list(set(conf['forcing_vars'].keys()).
+                        intersection(conf['constant_vars'].keys()))
+        [conf['forcing_vars'].pop(x) for x in del_keys]
 
     if conf['forcing_fmt'] != 'binary':
         conf['forcing_vars'] = invert_dict(conf['forcing_vars'])
--- a/metsim/cli/ms.py
+++ b/metsim/cli/ms.py
@@ -63,7 +63,14 @@ def init(opts):
     def to_list(s):
         return json.loads(s.replace("'", '"').split('#')[0])
 
+
+    conf['constant_vars'] = OrderedDict(config['constant_vars'])
     conf['forcing_vars'] = OrderedDict(config['forcing_vars'])
+    # remove constant_vars from forcings_vars
+    del_keys = list(set(conf['forcing_vars'].keys()).
+                    intersection(conf['constant_vars'].keys()))
+    [conf['forcing_vars'].pop(x) for x in del_keys]
+
     if conf['forcing_fmt'] != 'binary':
         conf['forcing_vars'] = invert_dict(conf['forcing_vars'])
     conf['domain_vars'] = invert_dict(OrderedDict(config['domain_vars']))
--- a/metsim/cli/ms.py
+++ b/metsim/cli/ms.py
@@ -64,18 +64,13 @@ def init(opts):
         return json.loads(s.replace("'", '"').split('#')[0])
 
     conf['forcing_vars'] = OrderedDict(config['forcing_vars'])
-    if 'constant_vars' in config:
-        conf['constant_vars'] = OrderedDict(config['constant_vars'])
-        # remove constant_vars from forcings_vars
-        del_keys = list(set(conf['forcing_vars'].keys()).
-                        intersection(conf['constant_vars'].keys()))
-        [conf['forcing_vars'].pop(x) for x in del_keys]
-
     if conf['forcing_fmt'] != 'binary':
         conf['forcing_vars'] = invert_dict(conf['forcing_vars'])
     conf['domain_vars'] = invert_dict(OrderedDict(config['domain_vars']))
     conf['state_vars'] = invert_dict(OrderedDict(config['state_vars']))
     conf['chunks'] = OrderedDict(config['chunks'])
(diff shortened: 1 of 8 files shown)


metsim/metsim.py
```diff
@@ -247,6 +249,15 @@ def met_data(self):
...
             self._met_data['lat'] = self.domain['lat']
             self._met_data['lon'] = self.domain['lon']
+
+            # process constant_vars
+            template_var = list(self.params.get('forcing_vars', None).values())[0]
+            constant_vars = self.params.get('constant_vars', None)
+            for var in constant_vars.keys():
+                constant_value = float(constant_vars[var])
+                self._met_data[var] = self._met_data[template_var].copy()
+                self._met_data[var] += -self._met_data[var] + constant_value
```
bartnijssen: Clunky - there must be a better way to set a `DataArray` to a constant value.

metsim/cli/ms.py
```diff
@@ -63,7 +63,14 @@ def invert_dict(d):
     def to_list(s):
         return json.loads(s.replace("'", '"').split('#')[0])
 
+
+    conf['constant_vars'] = OrderedDict(config['constant_vars'])
     conf['forcing_vars'] = OrderedDict(config['forcing_vars'])
+    # remove constant_vars from forcings_vars
+    del_keys = list(set(conf['forcing_vars'].keys()).
+                    intersection(conf['constant_vars'].keys()))
+    [conf['forcing_vars'].pop(x) for x in del_keys]
```
bartnijssen: This way the `constant_vars` are never even read even if they are also mentioned in `forcing_vars`

metsim/metsim.py
```diff
@@ -539,6 +550,12 @@ def _validate_setup(self):
         if not len(self.params.get('forcing', [])):
             errs.append("Requires input forcings to be specified")
 
+        # Make sure there is at least one forcing_var
+        # They cannot all be constant since we use one as a template
+        # for the others
+        if not len(self.params.get('forcing_vars', [])):
+            errs.append("Requires at least one non-constant forcing")
```
bartnijssen: This error never gets "exercised", because `_self_validate()` is executed *AFTER* the met data are read in. We should probably change that a little so we can check the setup *BEFORE* reading the met data (see also issue #169)

metsim/metsim.py
```diff
@@ -247,6 +249,15 @@ def met_data(self):
             self._met_data['elev'] = self.domain['elev']
             self._met_data['lat'] = self.domain['lat']
             self._met_data['lon'] = self.domain['lon']
+
+            # process constant_vars
+            template_var = list(self.params.get('forcing_vars', None).values())[0]
+            constant_vars = self.params.get('constant_vars', None)
+            for var in constant_vars.keys():
+                constant_value = float(constant_vars[var])
+                self._met_data[var] = self._met_data[template_var].copy()
```
bartnijssen: copying the `template_var` ensures that the constant field is of the right dimensions. The correct attributes get set when the variable is written to file

bartnijssen:
I'll make some more edits to make sure it passes the tests in case `constant_vars` is not defined.

metsim/metsim.py
jhamman: I think you want `xarray.full_like`: http://xarray.pydata.org/en/stable/generated/xarray.full_like.html#xarray.full_like

metsim/cli/ms.py
bartnijssen: Actually that does not work with the binary and ascii reads in which case you *must* read all the met variables. So nice but deleted in later update

bartnijssen:
@arbennett - the test for python3.5 fails (3.6 passes), which has nothing to do with this PR. Same for PR #172 .

metsim/metsim.py
bartnijssen: No longer used - now using `xr.full_like()`

bartnijssen:
Added tests - passes all tests on local machine (py 3.7 ) and CI py 3.6 test. Fails on CI py 3.5 test, which I think is unrelated to this PR (but maybe not).

metsim/metsim.py
```diff
@@ -247,6 +249,18 @@ def met_data(self):
             self._met_data['elev'] = self.domain['elev']
             self._met_data['lat'] = self.domain['lat']
             self._met_data['lon'] = self.domain['lon']
+
+            # process constant_vars
+            constant_vars = self.params.get('constant_vars', None)
+            if constant_vars:
+                if self.params['forcing_fmt'] != 'binary':
+                    template_var = list(self.params.get('forcing_vars', None).values())[0]
```
arbennett: I think it would be clearer, and eliminate the `if` statement to get the template variable by doing something like `template_var = self._met_data[list(self._met_data)[0]]`

metsim/metsim.py
```diff
@@ -247,6 +249,18 @@ def met_data(self):
...
+                else:
+                    template_var = list(self.params.get('forcing_vars', None).keys())[0]
+                for var in constant_vars.keys():
+                    self._met_data[var] = xr.full_like(self._met_data[template_var], 
+                                                       float(constant_vars[var]))
```
arbennett: I would prefer if we did the type-casting in `cli.py` so that it's clear that what should be in `constant_vars[var]` is a float. This makes it stricter for scripting usage, so that users can't specify their constant variables as strings.

metsim/metsim.py
bartnijssen: Done

metsim/metsim.py
bartnijssen: I would tend to agree, except that is not what is done in the rest of the code either. There are no typecasts in `cli.py` and no meaningful validation of paths and arguments is done before an attempt is made to read the data (resulting in uninformative error messages). 

Currently values are typecast at the time of use (`params['time_step']`) is cast to int repeatedly. So I would like to separate that, i.e. move the `validate` function up (see issue #169 for example) and fail early. 


``` python
 c = int(self.params['chunks'].get(d, s))
 self.disagg = int(self.params['time_step'])
```

bartnijssen: So I suggest not fixing this now

bartnijssen:
Still tracking down the failure of the test in 3.5

PR #186, opened 2019-10-15
Fix/dimension coordinate

### LINKED ISSUE #184: Add sequential coordinate variables for dimensions without coordinates
From @jhamman a description of an issue which can occur if there are dimensions that have no coordinate variables associated:

> Just to quickly share a bit of detail on what was going wrong under the hood. Xarray uses the notion of `Coordinates`, derived from the CF conventions, to describe the labels along axis (dimensions). It also support non-dimension coordinates to a limited extent (not all operations are possible when using multi-dimensional coordinates). A good description of these distinctions is here: http://xarray.pydata.org/en/stable/data-structures.html#coordinates

> There was an implicit assumption, pretty deep in Metsim, that these chunk dimension had a dimension-coordinate associated with the domain/forcing file. Because this was not true in @andywood’s setup, things got a little wacky when xarray assigned a (required) coordinate for the hru dimension.

> The fix in MetSim will require adding a default dimension coordinate for datasets that do not provide one. This fix is going to be particularly important for the case where the coordinates (lat/lon) are multidimensional 

And from myself:

> The input we've been giving MetSim has no coordinate variable for `hru` (Though we do have `hruId`, but MetSim does not know to use it) which causes MetSim to just keep selecting the same first CHUNKSIZE of the input forcings. Without getting too much in the details of how MetSim works, we can provide a temporary fix for your datasets and alongside this I'll work on a general solution.

arbennett:
- [x] closes #184 
 - [x] tests added / passed
 - [x] whatsnew entry


jhamman:
we need some sort of test here. Especially because some of the logic in this PR will be replaced when we move to the map_blocks solution for chunked execution.

PR #192, opened 2019-10-17
Add ability to shift output by one timestep

arbennett:
Re-reverting this because I got hasty on a previous pull request.

 - [ ] closes #xxxx
 - [ ] tests added / passed
 - [ ] whatsnew entry


jhamman:
LGTM, can we add a test that runs with this setting?

PR #201, opened 2020-02-07
Use args argument for parse_args

Chilipp:
This small fix makes sure that the `parse(args)` function uses the given `args`

 - [ ] tests added / passed
 - [ ] whatsnew entry


arbennett:
Looks good to me - thanks for the addition!

PR #220, opened 2020-05-18
Fix #219

### LINKED ISSUE #219: UTC offset is set to true even if set to false in configuration
It seems that the `utc_offset` flag in the configuration is getting set to `True` any time it is mentioned in the configuration file (regardless of value). The `utc_offset` can be made to `False` by excluding it from the configuration file.

arbennett:
- [x] closes #219
 - [x] tests added / passed
 - [x] whatsnew entry

### DIFF
--- a/metsim/cli/ms.py
+++ b/metsim/cli/ms.py
@@ -83,6 +83,12 @@ def init(opts):
     else:
         forcing_files = conf['forcing']
 
+    if 'utc_offset' in conf.keys() and conf['utc_offset'].strip() == 'True':
+        conf['utc_offset'] = True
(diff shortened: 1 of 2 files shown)


metsim/cli/ms.py
```diff
@@ -83,6 +83,12 @@ def to_list(s):
     else:
         forcing_files = conf['forcing']
 
+    if 'utc_offset' in conf.keys() and conf['utc_offset'].strip() == 'True':
```
bartnijssen, line 86: Can you change this to:
`conf['utc_offset'].strip() == 'true'`

PR #229, opened 2020-07-07
[3] Feature/yaml config

arbennett:
This PR allows for yaml as a configuration file format. See the updated docs or the example given for an example configuration with the new format.

 - [x] tests added / passed
 - [x] whatsnew entry


docs/examples.rst
```diff
@@ -7,80 +7,85 @@ Basics
...
+        # Format is metsim_name: input_name
+        prec  : 'Prec'
+        t_max : 'Tmax'
+        t_min : 'Tmin'
+
+    state_vars:
+        # Format is metsim_name: input_name
+        prec  : 'prec'
+        t_max : 't_max'
+        t_min : 't_min'
```
bartnijssen, line 66: This is a bit confusing. Why would the `state_vars` input names be different than the `forcing_vars` (I mean - you could I guess, but is that really the common use case?)

examples/example_yaml.yaml
```diff
@@ -0,0 +1,65 @@
...
+    # Format is metsim_name: input_name
+    prec  : 'Prec'
+    t_max : 'Tmax'
+    t_min : 'Tmin'
+
+state_vars:
+    # Format is metsim_name: input_name
+    prec  : 'prec'
+    t_max : 't_max'
+    t_min : 't_min'
```
bartnijssen, line 53: See earlier comment

docs/examples.rst
arbennett: I'll update this so that they're the same. I don't really know if it's common, but I do have at least a couple test cases where this is true, which is probably more of an artifact of MetSim changing through time rather than anything else.

